PR #67, opened 2014-08-14
Notification service implementation based on RabbitMQ

kamenim:
This is going to work in conjuction with Dovecot plugin for Openchange
and also notification consumer implemented in Openchange

So we have:
* Dovecot plugin for Openchange. Whenever new message arrives,
  it pushes a message in RabbitMQ
* Daemon (provided by this patch) listens for such RMQ messages
  and register them into mapistore indexing
* When message is registered we publish RMQ message into
  user specific queue
* TODO: A consumer in Openchage layer consumes such messages
  and notify Outlook for changes - not fully implemented yet.

NOTE: We should have RMQ message handler implemented into Openchange
      for this to work.


jkerihuel:
new design and implementation in the pipe - therefore closing

PR #71, opened 2014-08-18
provision: Add new schemas for Transport Settings and Accepted Domains

bencer:


jkerihuel:
Pending for more 7 months already. If it is not useful right now and if a rebase can't reasonably be expected this week, I will close it by the end of the week.

javierag:
I will mail the author asking about this. I vaguely remember that was neccesary fro a integration with a third party

javierag:
Moreover, if we change the schema we should add migration for already provisioned systems

javierag:
The author says that his is needed for compatibility with exchange but I need to get more details from him

jkerihuel:
Pending for too long already and no action taken which means it is not critical for openchange to move on. Closing PR. Feel free to open it again in the future once it is rebased.

javierag:
I agree with this resolution

PR #135, opened 2014-11-18
Not allow to provision where there is already one provisioned organization

javierag:
Also better assertion in guessing names to detect having more than one organization provisioned


jkerihuel:
@javierag: Can you update this PR to match latest master and @blaxter comments for easier integration?

javierag:
Sorry, this deprecated.
In master this is done via the check_not_provsioned method in openchange.provision

PR #156, opened 2014-11-30
Add hook to set 'dcerpc endpoint servers' by default.

jelmer:
This is mostly a RFC; this change depends on a matching change in Samba.

See https://lists.samba.org/archive/samba-technical/2014-November/103935.html

### DIFF
--- a/mapiproxy/dcesrv_mapiproxy.c
+++ b/mapiproxy/dcesrv_mapiproxy.c
@@ -782,11 +782,13 @@ NTSTATUS samba_init_module(void)
 	status = dcerpc_server_mapiproxy_init();
 	NT_STATUS_NOT_OK_RETURN(status);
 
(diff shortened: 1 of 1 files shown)


blaxter:
:+1: for me to merge, but I'd suggest to wait :hand: until we upgrade [samba version](https://github.com/openchange/openchange/blob/master/script/samba4_ver.sh) to one with this patch already included.

jelmer:
Updated. 

This hook is only used if the Samba that is used is new enough to support this hook (Samba >= 4.3). 

Adding the openchange dcerpc endpoint servers to the list of dcerpc endpoint servers will only actually enable them if openchange has also been provisioned. In other words, an admin that install openchange but does not set it up any other way will not automatically (and surprisingly) have OpenChange server enabled.

jkerihuel:
:+1: LGTM

PR #157, opened 2014-11-30
Change configuration defaults for mapiproxy.

jelmer:
Change configuration defaults for mapiproxy to:
    
dcerpc_mapiproxy:server = true
dcerpc_mapiproxy:interfaces = exchange_emsmdb, exchange_nsp, exchange_ds_rfr
    
Users can still go back to the old defaults by explicitly setting:
dcerpc_mapiproxy:server = false
dcerpc_mapiproxy:interfaces =

This reduce the amount of changes users need to make to their configuration when setting up OpenChange.

### DIFF
--- a/mapiproxy/dcesrv_mapiproxy.c
+++ b/mapiproxy/dcesrv_mapiproxy.c
@@ -605,7 +605,7 @@ static NTSTATUS mapiproxy_op_init_server(struct dcesrv_context *dce_ctx, const s
 {
 	NTSTATUS		ret;
 	struct dcesrv_interface	iface;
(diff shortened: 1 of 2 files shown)


blaxter:
I'd be more than happy to merge this now :+1: but `mapiproxy/libmapiproxy/dcesrv_mapiproxy_server.c:182` need some love as well (by love I mean s/false/true :monkey_face:), @jelmer could you add this change?

jelmer:
Nice catch - fixed.

blaxter:
:+1:

PR #162, opened 2014-12-06
Add proposal for openchange.conf.

jelmer:


jelmer:
any more thoughts on this, especially now that #156 has landed?

jkerihuel:
:+1: for me and the way to go for mid-term/long-term plans

PR #182, opened 2015-01-08
Framework to manage MySQL schema changes written in Python

sixstone-qq:
Features:

  * It provides migration indexed by a number
  * It can apply/unapply migrations to a defined version
  * It can list migrations applied or not
  * It can fake to a defined version

Added new command to manage it called `openchange_migrate`.

It applies to the provision of OpenChangeDB for MAPIProxy part and Indexing and Named Properties in their MySQL backends for MAPIStore section.

It is highly inspired by @blaxter and @exekias code along with Django migrations code.


python/openchange/migration/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        if current_version < target_version:
+            ran = range(current_version + 1, target_version + 1)
+            func = self.apply_version
+        elif current_version > target_version:
+            ran = range(target_version + 1, current_version + 1)
+            func = self.unapply_version
+        else:
+            return False
+
+        for ver in ran:
```
blaxter: maybe check if the migration `ver` exists? There is no constraint that migrations should have consecutive numbers as their version.

blaxter:
Could you also update `README.smbconf.md` file and remove `openchangedb:data` entry (no longer exists)?

Besides that looks really good to me.

mapiproxy/libmapiproxy/backends/openchangedb_mysql.c
```diff
@@ -3780,18 +3779,16 @@ enum MAPISTATUS openchangedb_mysql_initialize(TALLOC_CTX *mem_ctx,
...
-
-		if (!schema_created) {
-			DEBUG(1, ("Failed openchangedb schema creation, "
-				  "last mysql error was: `%s`\n", mysql_error(conn)));
+		migration_cmd = talloc_asprintf(mem_ctx, "openchange_migrate "
+						"--openchangedb-uri=%s openchangedb", connection_string);
+		OPENCHANGE_RETVAL_IF(!migration_cmd, MAPI_E_NOT_ENOUGH_MEMORY, NULL);
+		DEBUG(3, ("Creating schema for openchangedb on mysql %s with this command %s\n",
+			  connection_string, migration_cmd));
+		schema_created_ret = system(migration_cmd);
```
jkerihuel: I'm not OK with this system() approach.

sixstone-qq: What do you propose to run python code from C code?

jkerihuel: https://docs.python.org/2.5/ext/callingPython.html

We also have several references in mapistore-python branch where we call Python code from C

python/openchange/migration/__init__.py
sixstone-qq: Thanks for pointing out. Done :)

sixstone-qq:
migration framework is called from C using pure embedding Python.

python/openchange/mailbox.py
```diff
@@ -281,24 +288,93 @@ def set_message_ChangeNumber(self, server, ChangeNumber):
...
+        if not self.url.startswith('mysql://'):
+            raise ValueError("Bad connection string for mysql: invalid schema")
+
+        if '@' not in self.url:
+            raise ValueError("Bad connection string for mysql: expected format "
+                             "mysql://user[:passwd]@host/db_name")
+
+        if '/' not in self.url:
+            raise ValueError("Bad connection string for mysql: expected format "
+                             "mysql://user[:passwd]@host/db_name")
```
exekias: Why don't you use match return value to raise this kind of errors? if conn_string_match is None: raise ValueError

exekias:
:+1:

python/openchange/mailbox.py
sixstone-qq: Good catch :+1: , I'll do the checks after `conn_string_match` is `None`

blaxter:
:ok_hand:

icandigitbaby:
+1

papajulio:
:+1: Given by the automated tests, we found a crash of ocsmanager but @sixstone-qq already fixed in 2ee547f

jkerihuel:
LGTM

PR #183, opened 2015-01-10
Added indexes on indexing related tables

sixstone-qq:
Using migration framework from #182


exekias:
:+1:  to merge after #182

mapiproxy/util/schema_migration.c
```diff
@@ -0,0 +1,151 @@
...
+#include <util/debug.h>
+#include <Python.h>
+
+
+static PyObject *mailbox_module(void)
+{
+	PyObject *name = PyString_FromString("openchange.mailbox");
+	if (name == NULL)
+		return NULL;
+	return PyImport_Import(name);
```
jelmer: This might leak name ?

mapiproxy/util/schema_migration.c
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        if (PyTuple_SetItem(py_args, 0, PyString_FromString(connection_string)) == -1) {
+                PyErr_Print();
+                DEBUG(0, ("Unable to set first Python Tuple item\n"));
+                retval = -1;
+                goto end;
+        }
+
+        mailbox_obj = PyObject_CallObject(py_func, py_args);
```
jelmer: line 59 to line 82 could probably be replaced by a call to PyObject_CallMethod.

mapiproxy/util/schema_migration.c
```diff
@@ -0,0 +1,151 @@
+/*
+   Schema migration util procedures calling Python code
```
jelmer: This seems to be specific to migrating from SOGo to Mysql, is that correct?

python/openchange/mailbox.py
```diff
@@ -281,24 +288,93 @@ def set_message_ChangeNumber(self, server, ChangeNumber):
...
+
+        if '@' not in self.url:
+            raise ValueError("Bad connection string for mysql: expected format "
+                             "mysql://user[:passwd]@host/db_name")
+
+        if '/' not in self.url:
+            raise ValueError("Bad connection string for mysql: expected format "
+                             "mysql://user[:passwd]@host/db_name")
+
+        conn_string_match = re.match(r'mysql://(?P<user>.*?)(|:(?P<password>.*?))@(?P<host>.*?)(|:(?P<port>[0-9]+))/(?P<db>.*)', self.url)
```
jelmer: Please use the urlparse module (urlparse.urlparse) for URL parsing rather than regular expressions.

python/openchange/migration/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        cursor = self.db.cursor()
+        cursor.execute("""SELECT applied FROM {0}
+                          WHERE app = %s AND version = %s""".format(self.table_name),
+                       (app, version))
+        row = cursor.fetchone()
+        if row and row[0]:
+            return row[0]
+
+        return None
```
jelmer: "return None" is a noop :)

python/openchange/migration/indexing.py
```diff
@@ -0,0 +1,79 @@
...
+class InitialIndexingMigration(Migration):
+
+    description = 'initial'
+
+    @classmethod
+    def apply(cls, cur):
+        try:
+            cur.execute('SELECT COUNT(*) FROM `mapistore_indexing`')
+            return False
+        except:
```
jelmer: Please catch a more specific exception

python/openchange/migration/named_properties.py
```diff
@@ -0,0 +1,57 @@
...
+
+@migration('named_properties', 1)
+class InitialIndexingMigration(Migration):
+
+    description = 'initial'
+
+    @classmethod
+    def apply(cls, cur):
+        try:
+            cur.execute('SELECT COUNT(*) FROM `named_properties`')
```
jelmer: Please catch a more specific exception.

jelmer:
Mostly LGTM; added some minor comments.

It would be nice if the MySQL-specific code could be a bit more isolated to one part of the code rather than having "if url.startswith('mysql://')" spread everywhere. 

This just uses the term "migration" everywhere without specifying that this is the migration from SOGo to MySQL. At least the comments should make that clear, but perhaps the module names could also reflect it.

sixstone-qq:
@jelmer thank you very much for your comments. They are great and I'll fix them :). With regards to migration term, it means to migrate the "applications" within OpenChange (OpenChangeDB, Indexing and NamedProperties) the ones which has a MySQL backend. This has nothing to do with SOGo.

mapiproxy/util/schema_migration.c
sixstone-qq: No, have nothing to do with that. It is intended to run SQL operations to the following "applications" (OpenChangeDB, Indexing and NamedProperties) using python code

mapiproxy/util/schema_migration.c
sixstone-qq: Yup. Nice catch.

mapiproxy/util/schema_migration.c
sixstone-qq: :+1:

python/openchange/mailbox.py
sixstone-qq: Yes, good recommendation ;)

python/openchange/migration/__init__.py
sixstone-qq: Isn't it better to be explicit rather than implicit?

jelmer: Sorry, I missed that there was a code path in which it actually returned non-None. Disregard my comment, please :)

PR #185, opened 2015-01-12
Include modify_recipients

icandigitbaby:
* ` 54e7efd` includes small modifications in _rest.py_ that improve the semantics around `message_save` (e.g. folder permissions, restrictions, etc.). Message double-save is still not possible in OL.
* `6646bef` wraps in a method the way binary properties are encoded/decoded in order to push/retrieve them from/to the server
* `c52381e` includes various small fixes in _rest.py_
* `be58fac` implements the `RopModifyRecipients` operation in the REST API backend. The way recipients are pushed to the server (as a property) should be discussed. For instance, if we pass them as a property of the message dictionaries, maybe they could also be part of the message properties in _rest.py_ and not a separate attribute of message objects (in that case, `get_message_data` in _mapistore_python.c_ should change).
* `c5c34eb` modifies `emsmdbp_copy_message_recipients_mapistore` in _emsmdbp_object.c_: it used to modify the recipients only when the display name and the email address were not the first and second properties of the recipients. Now `modify_recipients` is independent of that condition (still dependent of `recipient_count > 0`).
* `54a22fe` modifies `get_message_data` in _mapistore_python.c_: It used to retrieve a fixed set of properties from the recipients. Now it includes all the properties in the recipients in the `mapistore_message` object. The method assumes that all the recipients have the same set of properties.

### DIFF
--- a/mapiproxy/libmapistore/backends/python/rest.py
+++ b/mapiproxy/libmapistore/backends/python/rest.py
@@ -142,10 +142,8 @@ class _RESTConn(object):
         self._dump_response(r)
         folders = r.json()
         newlist = []
-        for i,folder in enumerate(folders):
-            for key,value in folder.iteritems():
-                if mapistore.isPtypBinary(key):
-                    folder[key] = bytearray(base64.b64decode(str(value)))
+        for folder in folders:
+            self._decode_object_b64(folder)
             folder['PidTagInstID'] = folder['id']
             folder['PidTagInstanceNum'] = 0
             folder['PidTagRowType'] = 1
@@ -159,9 +157,7 @@ class _RESTConn(object):
         """
         r = self.so.get('%s%s' % (self.base_url, msg_uri))
         msg = r.json()
-        for key, value in msg.iteritems():
-            if mapistore.isPtypBinary(key):
-                msg[key] = bytearray(base64.b64decode(str(value)))
+        self._decode_object_b64(msg)
         return msg
 
     def get_message_count(self, uri):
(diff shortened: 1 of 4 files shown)


mapiproxy/libmapistore/backends/python/rest.py
```diff
@@ -519,10 +568,14 @@ def get_properties(self, properties):
...
-        properties['PidTagAccess'] = PidTagAccessFlag.Read|PidTagAccessFlag.HierarchyTable|PidTagAccessFlag.ContentsTable|PidTagAccessFlag.AssocContentsTable
+        properties['PidTagAccess'] = (PidTagAccessFlag.Modify |
+                                      PidTagAccessFlag.Read |
+                                      PidTagAccessFlag.HierarchyTable |
+                                      PidTagAccessFlag.ContentsTable |
+                                      PidTagAccessFlag.AssocContentsTable)
         properties['PidTagCreationTime'] = float((datetime.now(tz=timezone('Europe/Madrid')) - timedelta(hours=1)).strftime("%s.%f"))
         properties['PidTagLastModificationTime'] = properties['PidTagCreationTime']
-        properties["PidTagChangeKey"] = bytearray(uuid.uuid1().bytes + '\x00\x00\x00\x00\x00\x01')
+        properties["PidTagChangeKey"] = bytearray(uuid.UUID('082152cf-ae39-44a1-b11f-a1e1af3bac08').bytes + uuid.uuid1().bytes[0:6])
```
jkerihuel: This uuid is specific to your deployment. I also agree that the initial approach is not better, however, randomness for randomness I would like to keep the initial approach.

However add a /* FIXME: Generate PidTagChangeKey based on PidTagChangeNumber */

PR #189, opened 2015-01-15
Include submit_message

icandigitbaby:
This PR implements the submit operation for E-mails in _mapistore_python.c_ and _rest.py_. The latter doesn't check the return value of the server, which at this moment only takes the parameters (the message properties+recipients and the flag that indicates if any processing is needed).

The method doesn't check the permissions either, nor it modifies any properties apart from `PidTagClientSubmitTime`.


jkerihuel:
this PR is prior to the reset. pull -ff and recreate the PR

PR #190, opened 2015-01-15
Include submit_message

icandigitbaby:
This PR implements the submit operation for E-mails in mapistore_python.c and rest.py. The latter doesn't check the return value of the server, which at this moment only takes the parameters (the message properties+recipients and the flag that indicates if any processing is needed).

The method doesn't check the permissions either, nor it modifies any properties apart from `PidTagClientSubmitTime`.

### DIFF
--- a/mapiproxy/libmapistore/mapistore_python.c
+++ b/mapiproxy/libmapistore/mapistore_python.c
@@ -2617,6 +2617,70 @@ static enum mapistore_error mapistore_python_message_save(TALLOC_CTX *mem_ctx,
 	return retval;
 }
 
+/**
+   \details Submit message
+
+   \param mem_ctx pointer to the memory context
+   \param message_object pointer to the message object to submit
+   \param flags set of flags that determine the behaviour of the operation
+
+   \return MAPISTORE_SUCCESS on success, otherwise MAPISTORE error
+ */
+static enum mapistore_error mapistore_python_message_submit(void *message_object,
+							  enum SubmitFlags flags)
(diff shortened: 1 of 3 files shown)


python/mock/rest/api_server.py
```diff
@@ -388,7 +375,36 @@ def module_messages_get_attachments(msg_id):
...
 
 ###############################################################################
-# Calendar
+# Mail service
+###############################################################################
+
+@app.route('/mails/submit/', methods=['POST'])
+def module_mail_submit():
+    """ Send mail to its recipients.
+        param data['msg']: the message properties
```
sixstone-qq: You should use here param from sphinx but simply describe what JSON contains.

mapiproxy/libmapistore/mapistore_python.c
```diff
@@ -2617,6 +2617,70 @@ static enum mapistore_error mapistore_python_message_save(TALLOC_CTX *mem_ctx,
...
+   \return MAPISTORE_SUCCESS on success, otherwise MAPISTORE error
+ */
+static enum mapistore_error mapistore_python_message_submit(void *message_object,
+							  enum SubmitFlags flags)
+{
+	enum mapistore_error		    retval;
+	struct mapistore_python_object	*pyobj;
+	PyObject                        *message;
+	PyObject			            *pres;
+    PyObject                        *serv_spool;
```
sixstone-qq: You are mixing tabs and spaces here.

PR #191, opened 2015-01-16
Add check option to openchange_provision

javierag:
This option checks if the provision is possible. I added it because it is needed for Zentyal
server integration and I think it could be useful for more people.
If you do not find it useful, just clsoe the pull request and I will make a script
for Zentyal |server instead


blaxter:
:+1:

setup/openchange_provision
```diff
@@ -97,5 +98,7 @@ if not opts.openchangedb:
     else:
         parser.print_help()
         sys.exit(1)
+elif opts.check:
```
jelmer: what if both opts.check and opts.openchangedb are specified?

it would be nice to give users an error message in that case, otherwise the behaviour might be unexpected (especially considering the description doesn't say that --check is mutually exclusive with --openchangedb).

jelmer:
(Change itself and the code seems reasonable, but I thought this was worth mentioning)

setup/openchange_provision
javierag: It does the check operation..
I have fixed this and other similar instances in #199

PR #192, opened 2015-01-17
Remove default backend.

jelmer:
This means users will have to explicitly set *a* backend, whether it be ldb or mysql. 

I think this covers all code paths that deal with the default, but please let me know if there are others..


blaxter:
:+1:

PR #196, opened 2015-01-22
[mapistore-python]: Allow NULL properties in recipients

icandigitbaby:
This PR allows the modify_recipients and get_message_data functions in mapistore_python.c to process recipients with properties that aren't set. Thus, recipients in the REST API backend don't necessarily have the same set of properties (which was assumed before), so the code has been modified consequently.

### DIFF
--- a/mapiproxy/libmapistore/backends/python/sample.py
+++ b/mapiproxy/libmapistore/backends/python/sample.py
@@ -813,8 +813,13 @@ class MessageObject(BackendObject):
 
     def get_message_data(self):
         print '[PYTHON]: %s message.get_message_data()' % (self.name)
+        # Create a set with the union of the recipients' keys
+        recip_keys = set().union(*self.basedict["recipients"])
+        # Transform the set into a list
+        recip_keys = list(recip_keys)
 
-        return (self.basedict["recipients"], self.basedict["properties"])
+        return {'recipients': self.recipients, 'recip_keys': recip_keys,
+                'properties': self.properties}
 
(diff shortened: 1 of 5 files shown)


mapiproxy/libmapistore/backends/python/sample.py
```diff
@@ -813,6 +813,10 @@ def __init__(self, message, folder, mid, rw):
 
     def get_message_data(self):
         print '[PYTHON]: %s message.get_message_data()' % (self.name)
+        # Create a set with the union of the recipients' keys
+        keys = set().union(*self.basedict["recipients"])
+        # Transform the set into a list
+        keys = list(keys)
 
         return (self.basedict["recipients"], self.basedict["properties"])
```
sixstone-qq: You are not returning keys...

mapiproxy/libmapistore/mapistore_python.c
```diff
@@ -2646,34 +2616,33 @@ static enum mapistore_error mapistore_python_message_submit(void *message_object
...
+			serv_spool = Py_True;
+			break;
 	}
-
 	/* Call submit function */
 	pres = PyObject_CallMethod(message, "submit", "O", serv_spool);
 	if (pres == NULL) {
 		DEBUG(0, ("[ERR][%s][%s]: PyObject_CallMethod failed: ",
-			  pyobj->name, __location__));
+				pyobj->name, __location__));
```
sixstone-qq: These spaces changes are necessary?

PR #198, opened 2015-01-29
Relational operators in Table implementation

icandigitbaby:
This PR implements different relational operations used in restrictions, such as `NotEqual`, `LessThan`, etc. It doesn't implement the DL membership operator. Comparisons are made in a Pythonic way and types are not checked.

The PR also includes a small corrections that uses the wrapping function for decoding data from the REST calls in FAI message retrieval.

### DIFF
--- a/mapiproxy/libmapistore/backends/python/rest.py
+++ b/mapiproxy/libmapistore/backends/python/rest.py
@@ -210,10 +210,8 @@ class _RESTConn(object):
         self._dump_response(r)
         msgs = r.json()
         newlist = []
-        for i,msg in enumerate(msgs):
-            for key,value in msg.iteritems():
-                if mapistore.isPtypBinary(key):
-                    msg[key] = bytearray(base64.b64decode(str(value)))
+        for msg in msgs:
+            self._decode_object_b64(msg)
             msg['PidTagInstID'] = msg['id']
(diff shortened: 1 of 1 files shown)


jkerihuel:
I am going to put this PR on hold because I have pending commits that are doing a similar thing

mapiproxy/libmapistore/backends/python/rest.py
```diff
@@ -530,6 +528,19 @@ class PidTagAccessFlag(object):
     AssocContentsTable  = 0x00000020
 
 
+class RelOperator(object):
+    """ Define possible values for property comparison operators
```
jelmer: Please use PEP8: one-line summary for docstrings, no leading or trailing whitespace.

mapiproxy/libmapistore/backends/python/rest.py
```diff
@@ -530,6 +528,19 @@ class PidTagAccessFlag(object):
...
+        included in 'property' restriction structures """
+
+    LessThan            = 0x00
```
jelmer: Please use one space between name and equals sign, per PEP8.

icandigitbaby:
Relational operators were removed from the PR due to their implementation in a different one. The remaining commit is still valid, although the name of the PR doesn't apply anymore.

PR #199, opened 2015-02-02
Raise error when provision or newuser  commands are invoked with incompatible options

javierag:
Otherwise, as Jelmer Vernooij noticed, the triggered operation could be
a unexpected one.

### DIFF
--- a/setup/openchange_provision
+++ b/setup/openchange_provision
@@ -18,7 +18,7 @@
 #
 
 import optparse
-import os,sys
+import os, sys
 
 # To allow running from the source directory
 sys.path.append("python")
@@ -28,6 +28,24 @@ import openchange
 import samba
 import samba.getopt as options
 import openchange.provision as openchange
+
+def check_incompatible_options(parser, opts):
+    incompatible_options = ['deprovision', 'standalone', 'additional', 'primary_server', 'openchangedb', 'migrate', 'check']
+    active_option = None
+    for option_name in incompatible_options:
+        if getattr(opts, option_name, None):
+            if active_option:
+                if active_option == 'additional' and option_name == 'primary_server':
+                    # this combination is compatible
+                    continue
+                else:
+                    # for the error msg put the options in their command line shape
+                    active_option = '--' + active_option.replace('_', '-')                    
+                    option_name   = '--' + option_name.replace('_', '-')
+                    parser.error("The options " + active_option + " and " + option_name + " are mutually exclusive")
+            else:
+                active_option = option_name
+
 
 parser = optparse.OptionParser("openchange_provision [options]")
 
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks! Looks good to me.

I wonder though if it would make sense to use subcommands for these things (e.g. "openchange-provision deprovision") rather than options. Thoughts?

sixstone-qq:
@jelmer :+1: to your proposal.

setup/openchange_provision
```diff
@@ -29,6 +29,24 @@ import samba
 import samba.getopt as options
 import openchange.provision as openchange
 
+def check_incompatible_options(parser, opts):
+    incompatible_options = ['deprovision', 'standalone', 'additional', 'primary_server', 'openchangedb', 'migrate', 'check']
+    active_option = None
+    for option_name in incompatible_options:
+        if getattr(opts, option_name, None):
+            if active_option:
+                if (active_option == 'additional') and (option_name == 'primary_server'):
```
sixstone-qq: Parenthesis are useless here.

setup/openchange_provision
```diff
@@ -29,6 +29,24 @@ import samba
 import samba.getopt as options
 import openchange.provision as openchange
 
+def check_incompatible_options(parser, opts):
```
sixstone-qq: Why don't you use callbacks?
https://docs.python.org/2/library/optparse.html#callback-example-4-check-arbitrary-condition

setup/openchange_provision
```diff
@@ -29,6 +29,24 @@ import samba
...
+    incompatible_options = ['deprovision', 'standalone', 'additional', 'primary_server', 'openchangedb', 'migrate', 'check']
+    active_option = None
+    for option_name in incompatible_options:
+        if getattr(opts, option_name, None):
+            if active_option:
+                if (active_option == 'additional') and (option_name == 'primary_server'):
+                    # this combination is compatible
+                    continue
+                else:
+                    # for the error msg put the options in their command line shape
```
sixstone-qq: Using callbacks, this code is not longer necessary.

sixstone-qq:
@javierag there are two typos in the commit message:

"conttol" => "control"
"a unexpected one." => "an unexpected one" likewise "An unexpected journey"

javierag:
> I wonder though if it would make sense to use subcommands for these things (e.g. "openchange-provision deprovision") rather than options. Thoughts?

If they are always mutually exclusive, I think is better.  However this would broke anything which already uses this commands/

Anyway I do not see reason to revisit this. Maybe if we add more options in the future?.

setup/openchange_provision
javierag: They dont add anything there. In fact they would be more awkward since I will have to pass to the callback information about the values of other options

setup/openchange_provision
javierag: They help to mentally group things

sixstone-qq: They are redundant and avoid KISS principle.

setup/openchange_provision
blaxter: For me callbacks will make more sense, you check individually each parameter but if this works, I don't really care

setup/openchange_provision
blaxter: This is python, no parenthesis please.

blaxter:
To have subcommands is nice for tools with lots of behaviors (hello git, hello samba-tool), but this is a tool that you will use once and forget about it, so I don't think is a critical part, with this checks is enough, imho.

Anyway :+1: merging it!

jelmer:
I think even with a tool you're only going to use once or twice, it's important you can easily tell what modes it can operate in. With the current openchange provision tool, that is not possible. 

(I don' tmean to say revamping the structure of openchange_provision should be a part of this pull request)

PR #202, opened 2015-02-09
Fix returned value from `emsmdbp_object_open_folder_by_fid`

sixstone-qq:
From `enum mapistore_error` to `enum MAPISTATUS`.

Added an utility function to map MAPISTATUS to the mapistore_error as well to try to keep real error code available.

Fixed use case: 
Opening a shared calendar with no permissions returns MAPI_E_NO_ACCESS instead of MAPI_E_NOT_FOUND making work when the shared calendar access is later granted.


jkerihuel:
LGTM

papajulio:
:+1: Tests passed in Jenkins https://github.com/Zentyal/openchange/pull/119 :)

blaxter:
:+1:

PR #203, opened 2015-02-09
Correct set_read_flag behaviour

icandigitbaby:
This PR corrects the behaviour of `mapistore_python_message_set_read_flag`.

* The flags that the function takes define de behaviour of the function (set/clear read flag and send read receipt).
* The `PidTagMessageFlags` property is read and not overwritten.
* Read receipts are still unhandled after this PR.

### DIFF
--- a/mapiproxy/libmapistore/mapistore_python.c
+++ b/mapiproxy/libmapistore/mapistore_python.c
@@ -2652,7 +2652,7 @@ static enum mapistore_error mapistore_python_message_submit(void *message_object
 
    \param message_object pointer to the message object on which we
    want to set the property
-   \param flags the flags to set on the message
+   \param flags the flags that set the behaviour of the function
 
    \return MAPISTORE_SUCCESS on success, otherwise MAPISTORE_ERROR
  */
@@ -2662,10 +2662,13 @@ static enum mapistore_error mapistore_python_message_set_read_flag(void *message
 	enum mapistore_error		retval;
 	struct mapistore_python_object	*pyobj;
 	PyObject			*message;
+	PyObject                        *proplist;
 	PyObject			*pres;
 	PyObject			*key;
 	PyObject			*val;
 	PyObject			*pydict;
+	uint32_t                        msg_flags;
+	int                             ret;
 
 	/* Sanity checks */
 	MAPISTORE_RETVAL_IF(!message_object, MAPISTORE_ERR_INVALID_PARAMETER, NULL);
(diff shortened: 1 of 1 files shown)


mapiproxy/libmapistore/mapistore_python.c
```diff
@@ -2681,24 +2684,68 @@ static enum mapistore_error mapistore_python_message_set_read_flag(void *message
...
 
-	/* Set PidTagMessageFlags */
-	pydict = PyDict_New();
-	if (pydict == NULL) {
-		DEBUG(0, ("[ERR][%s]: Unable to initialize Python Dictionary\n", __location__));
+	/* Retrieve the message dictionary and read PidTagMessageFlags */
+	proplist = PyList_New(0);
+	if (proplist == NULL) {
+		DEBUG(0, ("[ERR][%s][%s]: Unable to initialize Python list\n",
+			  pyobj->name, __location__));
```
sixstone-qq: Why don't you call `PyErr_Print()` here?

icandigitbaby: Missed it, thanks!

mapiproxy/libmapistore/mapistore_python.c
```diff
@@ -2681,24 +2684,70 @@ static enum mapistore_error mapistore_python_message_set_read_flag(void *message
...
-	if (PyDict_SetItem(pydict, key, val) == -1) {
-		DEBUG(0, ("[ERR][%s]: Unable to add entry to Python dictionary\n",
+	val = PyDict_GetItem(pres, key);
+	if (val == NULL) {
+		/* If the property is missing, set all flags to zero */
+		msg_flags = 0x0;
+	} else {
+		msg_flags = PyLong_AsUnsignedLong(val);
+	}
+	Py_DECREF(pres);
```
jelmer: This seems to lack a call to Py_XDECREF(val)

mapiproxy/libmapistore/mapistore_python.c
```diff
@@ -2708,8 +2757,6 @@ static enum mapistore_error mapistore_python_message_set_read_flag(void *message
 		return MAPISTORE_ERR_CONTEXT_FAILED;
 	}
 
-	/* TODO Send Receipt */
```
jelmer: You probably want to check the type of pres here, because PyLong_AsLong() will not return anything predictable otherwise.

mapiproxy/libmapistore/mapistore_python.c
icandigitbaby: `PyDict_GetItem()` returns a borrowed reference, so the reference count of `val` it isn't our responsibility.

mapiproxy/libmapistore/mapistore_python.c
icandigitbaby: Good point, although at the moment we rely on the backend to return a mapistore error code. Since this happens in several methods, this change will affect the whole file, and we'll leave that for a future PR.

mapiproxy/libmapistore/mapistore_python.c
jelmer: Whoops, you're right. Sorry.

mapiproxy/libmapistore/mapistore_python.c
jelmer: Fair enough.

jelmer:
:+1:

jkerihuel:
:+1:

PR #204, opened 2015-02-09
Ship private copy of dlinklist.h from Samba.

jelmer:
This is a first step towards being able to build without having Samba installed.


blaxter:
:+1: road to a world without dependencies!

PR #206, opened 2015-02-09
Remove unnecessary includes of core/ files.

jelmer:


blaxter:
:+1:

    If it compiles, it works.
                      -- Dalai Lama

PR #209, opened 2015-02-10
openchangedb: create_folder must use the given change number

blaxter:
Now a new change number was being generated and the given one was ignored.


jelmer:
:+1:

PR #211, opened 2015-02-10
Use talloc_new instead of talloc_zero(NULL, TALLOC_CTX)

sixstone-qq:
As suggested by @blaxter and @jelmer to avoid creating two
talloc contexts where is not required to create a temporary
context.


blaxter:
:+1:

PR #212, opened 2015-02-12
Implement read backwards on a oxctabl in QueryRows ROP

sixstone-qq:
Instead of crashing.

And we have implemented in libmapi for MAPI clients:

* Give support to set read direction on QueryRows oxctabl in libmapi
* Current limitation: Read backwards from BOOKMARK_BEGINNING will always return the first row. MAPI clients should check Origin response field to know nothing else to read. Current `libmapi:QueryRows` does not support it.

### DIFF
--- a/utils/mapitest/modules/module_zentyal.c
+++ b/utils/mapitest/modules/module_zentyal.c
@@ -485,3 +485,105 @@ _PUBLIC_ bool mapitest_zentyal_4872(struct mapitest *mt)
 
 	return true;
 }
+
+/**
+    \details Test #6723 QueryRows with reverse read.
+
+    It is identified as CR-103 (https://projects.zentyal.com/issues/1543)
+
+    \param mt pointer to the top level mapitest structure
+
+    \return true on success, otherwise false
+*/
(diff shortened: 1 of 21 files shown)


utils/mapitest/modules/module_zentyal.c
```diff
@@ -485,3 +485,105 @@ _PUBLIC_ bool mapitest_zentyal_4872(struct mapitest *mt)
...
+		mapitest_print_retval(mt, "SetColumns");
+		return false;
+	}
+
+	/* Step 6. QueryRows backwards on TIS folder hierarchy */
+	idx = count - 1;
+	test_result = true;
+	seek_count = 0;
+	do {
+		retval = QueryRows(&obj_htable, 0x1, TBL_ADVANCE, TBL_FORWARD_READ, &SRowSet);
```
kamenim: perhaps this call should use TBL_BACKWARD_READ flag (as comment says)?

sixstone-qq: Bad night and undo operation :) nice catch!

blaxter:
:+1:

PR #214, opened 2015-02-15
Implement oc_log, as specified in the spec.

jelmer:
This also converts libmapi/ to use the new logging API.


libmapi/oc_log.c
```diff
@@ -0,0 +1,71 @@
...
+
+	nwritten = vsnprintf(line, sizeof(line), fmt_string, ap);
+
+	if (level <= 0) {
+		samba_level = -level;
+	} else {
+		samba_level = 0;
+	}
+
+	DEBUG(samba_level, ("%s\n", line));
```
sixstone-qq: I'd like to let the developer to insert the new line or not in the `fmt_string` as `DEBUG` macro from samba does.

jelmer:
@sixstone-qq Thanks, that's a good point. Updated now to add the trailing newline *only* if it is not yet present, consistent with the header comment and the spec.

blaxter:
I'm a bit confuse about the use of `OC_DEBUG` and `oc_log`. Should we use both of them?

jelmer:
Yes, one is for messages meant for users (and developers). The other is just for developers, and logs filename/line no.

This distinction matters, in particular for things like syslog - where we don't want to spam.

On 23 February 2015 11:32:57 am GMT+02:00, "Jesús García Sáez" <notifications@github.com> wrote:
>I'm a bit confuse about the use of `OC_DEBUG` and `oc_log`. Should we
>use both of them?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/openchange/openchange/pull/214#issuecomment-75511880

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

javierag:
How the oc_log is enabled or disabled? flag in the ./configure?

blaxter:
@jelmer mmm, from my pov the one just for developers is `log level = debug` and the other ones (log level > debug) are for users (and developers).

The filename/line thing could be just an option (similar to `debug timestamps` in samba).

@javierag you could disabled them putting log level high enough. This is not the final implementation, just stage 0 (see doc/developer/logging.txt).

jelmer:
@javerag oc_log currently just logs to the samba logs files and uses the samba configuration options (per the spec). Once that changes and we start writing to /var/log/openchange.log, we'll need to have openchange-specific configuration for logging.

jelmer:
@blaxter and that is still the case; OC_DEBUG() is a convenience wrapper that logs to OC_LOG_DEBUG+level by default, and includes the file/lineno by default.

blaxter:
My point is that I don't like to have two ways to log messages. For me it makes more sense to have an option to decide whether you want to include or not file/lineno on all messages than decide it in each log message.

jkerihuel:
@blaxter too many options kill the choice. Better to be simple, consistent and agree on one single format.  

For me :+1: for Jelmer's proposal which makes a step in the right direction of what we are aiming at.

blaxter:
@jkerihuel  I don't understand it, with jelmer's proposal you have two options everytime you want to log somethig: either to use `oc_log` or `OC_DEBUG`. My proposal is to have only one function/macro and then to have an option (in the future, no needed right now because we still rely on samba's `DEBUG` macro) like the one samba has `debug timestamps`.

jkerihuel:
@blaxter Jelmer's proposal is to have an incremental approach with OC_DEBUG in place of DEBUG to restrict scope of changes at once.

blaxter:
Also right now (we are still using `DEBUG` macro from samba) using `OC_DEBUG` we will have a log file (samba running as daemon) with duplicate information, example:

    [2015/02/24 02:29:45.867144,  3] ../source4/ntvfs/ntvfs_base.c:79(ntvfs_register)
      ../source4/ntvfs/ntvfs_base.c:79(ntvfs_register) NTVFS backend 'cifs' for type 1 registered
    [2015/02/24 02:29:45.867173,  3] ../source4/ntvfs/ntvfs_base.c:79(ntvfs_register)
      ../source4/ntvfs/ntvfs_base.c:79(ntvfs_register) NTVFS backend 'smb2' for type 1 registered

So my proposal is (1) to remove all calls `oc_log`, (2) use always `OC_DEBUG` but remove the additional `__location__` and `__pretty_function__` from the macro (and in the future when we get rid of samba's DEBUG macro we can decide to either include always that or whatever)

jelmer:
I don't think we should to call something OC_DEBUG() that is writing regular user entries, and not just debug info. In that case, let's get rid of the OC_DEBUG() macro instead.

FWIW Both of these functions were documented in the spec, so I'm a bit surprised to get this feedback about OC_DEBUG/oc_log now rather than back then.

You make a good point about __location__ and __pretty_function__; unfortunately the location reported by Samba's DEBUG() will always be in oc_log.c. So we'll either have to live with two locations temporarily or just use a #define for oc_log() until we have our own logging infrastructure.

blaxter:
@jelmer from the spec I understood that the usage, as a developer, would be to use `OC_DEBUG` and that's it.

The conflict that we have here is about the usage. Correct me if I wrong but you want to use `oc_log` for messages useful for the user (the administrator of the server) and to use `OC_DEBUG` for messages useful to the developer or advanced user.

But for doing that (imho) you already have log levels.

jelmer:
@blaxter We do already have log levels, but the problem with log levels is that there is no consistency as to how they are used. This is why I've added the enum - the log levels further in the range can be used freely for debugging stuff.

blaxter:
So the real problem is how the developers (including myself) write log messages. Which right now I agree is a mess and probably most of all `DEBUG` messages have wrong log level.

jelmer:
Right, so the idea is that having named log levels makes it easier to decide what to use.

I will update the header to clarify some of the ideas.

blaxter:
Ok, it's an unsual idea but I'm open to new ideas if you think is better this way. I'd like to heard about @jkerihuel and other devs opinions though.

jelmer:
I've updated the comments in oc_log.h to better explain the different types of logging.

jelmer:
ping?

jkerihuel:
@jelmer: I don't see a way to control how verbose is the output. From my PoV, -3,-2, -1 and 0 are no brainer, but how do we control which `OC_LOG` messages show up?

Next steps (after this merge) are presumably:
# Agree on `developers` internal log levels
# Agree on message prefix and unify source code once for all

jelmer:
For the moment, the log level is still determined by samba's "log level" setting since we use DEBUG. We'll have to add a setting once we migrate to our own log writer.

jelmer:
I think the developer internal log levels are kind of free form, like in Samba. The main thing that matters is that we have consistent logging levels for users IMHO. We could standardise on debug log levels for developers, but I'd rather keep that separate from this pull request.

jkerihuel:
ahh right.

Therefore :+1: for me as it stands

libmapi/oc_log.c
```diff
@@ -0,0 +1,74 @@
...
+	va_end(ap);
+}
+
+void oc_logv(enum oc_log_level level, const char *fmt_string, va_list ap)
+{
+	char line[OC_LOG_MAX_LINE];
+	int samba_level;
+	int nwritten;
+
+	nwritten = vsnprintf(line, sizeof(line), fmt_string, ap);
```
blaxter: unused variable, either check the returned value or remove it?

jkerihuel:
```
if (nwritten < 0) return;
```

jelmer: Done.

PR #215, opened 2015-02-15
Don't fail if setup/mapistore has already been removed.

jelmer:


sixstone-qq:
:+1:

jelmer:
Thanks for the review :)

PR #217, opened 2015-02-19
In get_AppointmentRecurrencePattern ignore reserved block length fields ...

javierag:
...if not present in the pattern.

This should avoid crashes with clients that send the message without caring about this fields.
However if they are exceptions to the recurrence we cannot do so because the ExtendedException array is just after the ReservedBlock1. In that case we continue to return error.

### DIFF
--- a/ndr_mapi.c
+++ b/ndr_mapi.c
@@ -2501,31 +2501,43 @@ _PUBLIC_ enum ndr_err_code ndr_pull_AppointmentRecurrencePattern(struct ndr_pull
 				NDR_CHECK(ndr_pull_ExceptionInfo(ndr, NDR_SCALARS, &r->ExceptionInfo[cntr_ExceptionInfo_0]));
 			}
 			NDR_PULL_SET_MEM_CTX(ndr, _mem_save_ExceptionInfo_0, 0);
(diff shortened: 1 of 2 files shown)


sixstone-qq:
:+1: , specially having unit tests along with it :)

blaxter:
Message commit not following guidelines

javierag:
Changed message

ndr_mapi.c
```diff
@@ -2501,31 +2501,42 @@ _PUBLIC_ enum ndr_err_code ndr_pull_AppointmentRecurrencePattern(struct ndr_pull
...
+				}
+				NDR_PULL_SET_MEM_CTX(ndr, _mem_save_ReservedBlock2_0, 0);
 			}
-			NDR_PULL_SET_MEM_CTX(ndr, _mem_save_ReservedBlock2_0, 0);
+
 			NDR_CHECK(ndr_pull_trailer_align(ndr, 4));
+		} else {
+			/* If there are extended exceptions we have missing data */
+			if (r->ExceptionCount > 0) {
+				return NDR_ERR_INCOMPLETE_BUFFER;
```
blaxter: undefined error for current samba version. We could return another error and put a FIXME comment?

PR #222, opened 2015-03-03
mapiproxy: single samdb connection

blaxter:
There is no need to create a new one for each bound session. So now all nspi and emsmdb session will use the same ldb_context. This will fix a problem (when `dcerpc_mapiproxy:samdb_url` is defined and not local) of too many ldap connection opened.

Also in case the samdb is not local (`dcerpc_mapiproxy:samdb_url` defined) the samdb context is open with reconnect flag.

I'm also introducing a new **CHANGES.md** file which should include for now on (if we see is useful) the new changes from an user perspective.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,18 @@
+# Changes
+
+All notable changes to this project will be documented in this file.
+The descriptions should be useful and understandable for end users of OpenChange.
+Unreleased changes refer to our current [master branch](https://github.com/openchange/openchange/).
+
+## [Unreleased]
+
+### Fixes
+* Fixed `Too many connections to ldap` when openchange runs on samba as member of a domain.
+
+
+
+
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -0,0 +1,18 @@
...
+
+## [Unreleased]
+
+### Fixes
+* Fixed `Too many connections to ldap` when openchange runs on samba as member of a domain.
+
+
+
+
+[//]: # (unreleased compare link should be changed to the lastest release)
```
blaxter: This is a way to include comments on markdown files

sixstone-qq: Typo: latest

exekias:
:+1: to both changes

jkerihuel:
:+1:

jkerihuel:
:warning: Check https://github.com/openchange/openchange/pull/225

PR #225, opened 2015-03-06
mapiproxy: do not release samdb connection

blaxter:
Fix ad02428d3b504380e6ddb90df242e1ff537048ca where samdb connections were changed to have only one. We cannot use the memory context of the emsmdb/nspi session but use a global one so they won't be released until the end.


jkerihuel:
:+1:

jkerihuel:
@blaxter :hand: :warning:: This code still has one major and insidious defect. `samdb_connect` and `samdb_connect_url` are registering samdb_ctx within tevent event loop which has been allocated with `mem_ctx` of the session.

When the session is released, then so is the event loop, hence leave an allocated samdb_ctx pointer but relying on an invalid `ev`. This lead to the crash of openchange server in emsabp.c :bug:.

PR #208, opened 2015-02-10
Fixed NSPI UpdateStat and QueryRows methods

javierag:
Also added to NSPI methods the codepage check mandated by specifications

The NSPI UpdateStat and QueryRow method are called when browsing the entries in the address book before they failed in their implemntation and if you had more of 40 entries it dont worked correctly.

PR #230, opened 2015-03-13
nspi: update stat fix

blaxter:
Regression on #208 I read so many times that pull request but I miss this, the only real change is  4f6be72


sixstone-qq:
In order to avoid this problem in the future, could we add unit test to this section or do something else?

javierag:
:+1: 

I missed it, if you don't click in the address list filter combobox, this path is not taken

PR #236, opened 2015-03-14
Optional pass of source dirs to gdb in debugoc.py

jacalvo:
Useful to run it from a wrapper script with a list of dirs as commandline args.


jkerihuel:
:+1:

PR #237, opened 2015-03-16
nspi: Added missing call in dcesrv_NspiUpdateStat PR#208

javierag:
I lost this method call in the refactoring of #208  . Also this path is not taken if you not select a filter in the addresses dialog display so i not detected it in the smoke test.


blaxter:
dupe #230

PR #238, opened 2015-03-16
openchange_user_cleanup: parse imap config properly

blaxter:
We can have both `imap://host:port` and `host:port` on sogo config so the regex should have the scheme as an optional match.


sixstone-qq:
:+1:

jkerihuel:
LGTM :+1:

PR #239, opened 2015-03-17
Added NSPI support for group accounts

javierag:


javierag:
i ahv to arange abit this

PR #240, opened 2015-03-17
Added NSPI support for group accounts

javierag:
- Groups appear now in addresses list
- Added directory attributes to change icons from groups in addresses lists
- Added openchange_group command to enable groups for openchange
- Added needed migration from directory attributes

However the groups need this patch to be actually be able to receive mail https://github.com/Zentyal/sogo/pull/100

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1052,6 +1052,7 @@ provision-install: python-install
 	$(INSTALL) -m 0755 setup/openchange_migrate $(DESTDIR)$(sbindir)/
 	$(INSTALL) -m 0755 setup/openchange_newuser $(DESTDIR)$(sbindir)/
 	$(INSTALL) -m 0755 setup/openchange_neworganization $(DESTDIR)$(sbindir)/
+	$(INSTALL) -m 0755 setup/openchange_group $(DESTDIR)$(sbindir)/
 	$(INSTALL) -d $(DESTDIR)$(samba_setupdir)/AD
 	$(INSTALL) -m 0644 setup/AD/oc_provision* $(DESTDIR)$(samba_setupdir)/AD/
 	$(INSTALL) -m 0644 setup/AD/prefixMap.txt $(DESTDIR)$(samba_setupdir)/AD/
--- a/mapiproxy/servers/default/nspi/emsabp.c
+++ b/mapiproxy/servers/default/nspi/emsabp.c
@@ -521,14 +521,6 @@ _PUBLIC_ void *emsabp_query(TALLOC_CTX *mem_ctx, struct emsabp_context *emsabp_c
 	case PR_ADDRTYPE_UNICODE:
 		data = (void *) talloc_strdup(mem_ctx, EMSABP_ADDRTYPE /* "SMTP" */);
 		return data;
-	case PR_OBJECT_TYPE:
-		data = talloc_zero(mem_ctx, uint32_t);
-		*((uint32_t *)data) = MAPI_MAILUSER;
-		return data;
-	case PR_DISPLAY_TYPE:
-		data = talloc_zero(mem_ctx, uint32_t);
-		*((uint32_t *)data) = DT_MAILUSER;
-		return data;
 	case PR_SEND_RICH_INFO:
 		data = talloc_zero(mem_ctx, uint8_t);
 		*((uint8_t *)data) = false;
@@ -632,6 +624,13 @@ _PUBLIC_ void *emsabp_query(TALLOC_CTX *mem_ctx, struct emsabp_context *emsabp_c
 			mvszA->lppszA[i] = talloc_strdup(mem_ctx, (char *)ldb_element->values[i].data);
 		}
 		data = (void *) mvszA;
+		break;
+	case PT_LONG:
+		ldb_val = ldb_msg_find_ldb_val(msg2, attribute);
+		if (!ldb_val) return NULL;
+		/*  There is duplication of code between ldb_msg_find_ldb_val and ldb_msg_find_attr_as_uint but we cannot afford to not discriminate between null value and default 0 value */
+		data = talloc_zero(mem_ctx, uint32_t);
+		*((uint32_t *)data) = ldb_msg_find_attr_as_uint(msg2, attribute, 0);
 		break;
 	default:
 		OC_DEBUG(3, "Unsupported property type: 0x%x", (ulPropTag & 0xFFFF));
@@ -863,8 +862,8 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				lpProps.value.l = 0x0;
 				break;
 			case PidTagDisplayName:
-				/* This value is temporal to workaround PropertyRow_addprop internals */
-				/* It will be set to NULL aftr the call to PropertyRow_addprop */
+				/* FIXME: This value is temporal to workaround PropertyRow_addprop internals */
+				/* It will be set to NULL after the call to PropertyRow_addprop */
 				lpProps.value.lpszW = "t";
 				break;
 			case PR_EMS_AB_IS_MASTER:
@@ -874,7 +873,7 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				break;
 			}
 			PropertyRow_addprop(aRow, lpProps);
-			/* PropertyRow_addprop internals overwrite with MAPI_E_NOT_FOUND when data is NULL */
+			/* FIXME: PropertyRow_addprop internals overwrite with MAPI_E_NOT_FOUND when data is NULL */
 			if (SPropTagArray->aulPropTag[i] == PR_DISPLAY_NAME ||
 			    SPropTagArray->aulPropTag[i] == PR_DISPLAY_NAME_UNICODE) {
 				aRow->lpProps[aRow->cValues - 1].value.lpszA = NULL;
--- a/mapiproxy/servers/default/nspi/emsabp.c
+++ b/mapiproxy/servers/default/nspi/emsabp.c
@@ -863,7 +863,9 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				lpProps.value.l = 0x0;
 				break;
 			case PidTagDisplayName:
-				lpProps.value.lpszW = NULL;
+				/* This value is temporal to workaround PropertyRow_addprop internals */
+				/* It will be set to NULL aftr the call to PropertyRow_addprop */
+				lpProps.value.lpszW = "t";
 				break;
 			case PR_EMS_AB_IS_MASTER:
 				lpProps.value.b = false;
@@ -889,23 +891,14 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				emsabp_PermanentEntryID_to_Binary_r(mem_ctx, permEntryID, &(lpProps.value.bin));
 				break;
 			case PR_CONTAINER_FLAGS:
-			  switch ((int)child) {
-				case true:
-					lpProps.value.l = AB_RECIPIENTS | AB_SUBCONTAINERS | AB_UNMODIFIABLE;
-					break;
-				case false:
+				if (child) {
 					lpProps.value.l = AB_RECIPIENTS | AB_UNMODIFIABLE;
+				} else {
+					lpProps.value.l = AB_RECIPIENTS | AB_SUBCONTAINERS | AB_UNMODIFIABLE;
 				}
 				break;
 			case PR_DEPTH:
-			  switch ((int)child) {
-				case true:
-					lpProps.value.l = 0x1;
-					break;
-				case false:
-					lpProps.value.l = 0x0;
-					break;
-				}
+				lpProps.value.l = child ? 0x1 : 0x0;
 				break;
 			case PR_EMS_AB_CONTAINERID:
 				dn = ldb_msg_find_attr_as_string(msg, "distinguishedName", NULL);
--- a/python/openchange/mailbox.py
+++ b/python/openchange/mailbox.py
@@ -23,7 +23,7 @@
 import os
 import os.path
 import samba
-from samba import Ldb, unix2nttime
+from samba import Ldb
 import ldb
 from urlparse import urlparse
 import uuid
@@ -550,3 +550,30 @@ class NamedPropertiesWithMysqlBackend(MysqlBackendMixin):
         self.url = url
         self._connect_to_mysql()
         self.migration_app = 'named_properties'
+
+
+class Directory(MysqlBackendMixin):
+    """
+    The directory itself does not use mysql but we have this class for ease the migrations
+    """
+
+    def __init__(self, url):
+        self.url = url
+        self._connect_to_mysql()
+        self.migration_app = 'directory'
+
+    def migrate(self, version=None, **kwargs):
+        """Migrate directory schema and data
+
+        :param int version: indicating which version to migrate. None migrates
+                            to the lastest version.
+        :param named lp: samba Loadparm context
+        :param named cred: samba credentials
+        :param named names: samba provision name
+        :param named setup_path: method to get the setup ldif files path
+        :rtype bool:
+        :returns: True if any migration has been performed
+        """
+        self.db.select_db(self.db_name)
+        migration_manager = MigrationManager(self.db, self.db_name)
+        return migration_manager.migrate(self.migration_app, version, **kwargs)
--- a/python/openchange/migration/__init__.py
+++ b/python/openchange/migration/__init__.py
@@ -118,7 +118,7 @@ class MigrationManager(object):
         self.db.commit()
         self._version[app] = None  # To reload the value
 
-    def migrate(self, app, target_version=None):
+    def migrate(self, app, target_version=None, **kwargs):
         """
         Apply/unapply missing migrations and update version information.
 
@@ -126,6 +126,7 @@ class MigrationManager(object):
 
         :param str app: the "application" to migrate to the target version
         :param int target_version: the target version to apply/unapply migrations.
+        :param dict kwargs: keyword parameters needed for migration if any
         :rtype bool:
         :returns: True if any migration was applied/unapplied
         """
@@ -148,9 +149,9 @@ class MigrationManager(object):
             # Migrate inside a transaction
             cursor = self.db.cursor()
             if go_forward:
-                migration.apply(cursor)
+                migration.apply(cursor, **kwargs)
             else:
-                migration.unapply(cursor)
+                migration.unapply(cursor, **kwargs)
             self.db.commit()
             cursor.close()
             if go_forward:
@@ -267,7 +268,7 @@ class Migration(object):
     description = 'migration'
 
     @classmethod
-    def apply(cls, cur):
+    def apply(cls, cur, **kwargs):
         """Apply this migration
 
         :param cur: the DB cursor
@@ -284,4 +285,4 @@ class Migration(object):
 
 
 # Import to get the migrations from different applications
-from . import indexing, openchangedb, named_properties
+from . import directory, indexing, openchangedb, named_properties
--- a/python/openchange/migration/directory.py
+++ b/python/openchange/migration/directory.py
@@ -1,0 +1,163 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# OpenChangeDB migration for directory schema and contents
+# Copyright (C) Javier Amor Garcia <jamor@zentyal.com> 2015
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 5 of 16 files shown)


Makefile
```diff
@@ -1052,6 +1052,7 @@ provision-install: python-install
 	$(INSTALL) -m 0755 setup/openchange_migrate $(DESTDIR)$(sbindir)/
 	$(INSTALL) -m 0755 setup/openchange_newuser $(DESTDIR)$(sbindir)/
 	$(INSTALL) -m 0755 setup/openchange_neworganization $(DESTDIR)$(sbindir)/
+	$(INSTALL) -m 0755 setup/openchange_group $(DESTDIR)$(sbindir)/
```
sixstone-qq: This requires to be included in zentyal-openchange packaging. So this file should be included here: https://github.com/Zentyal/openchange/tree/trusty/debian

mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.h
```diff
@@ -95,6 +95,7 @@ struct EphemeralEntryID {
 
 #define	EMSABP_DN	"/guid=%08X%04X%04X%02X%02X%02X%02X%02X%02X%02X%02X"
 #define	EMSABP_ADDRTYPE	"EX"
+//#define	EMSABP_ADDRTYPE	"SMTP"
```
sixstone-qq: Could you remove this line? I don't see the point of adding this commented code.

javierag: Thanks for the catch

Makefile
javierag: What is the procedure for this?. Should I create a branch or ask Julio when he does the merge?

sixstone-qq: Create a branch from trusty and do the Pull Request there, yes.

mapiproxy/servers/default/nspi/emsabp.c
```diff
@@ -863,7 +862,9 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				lpProps.value.l = 0x0;
 				break;
 			case PidTagDisplayName:
-				lpProps.value.lpszW = NULL;
+				/* This value is temporal to workaround PropertyRow_addprop internals */
+				/* It will be set to NULL aftr the call to PropertyRow_addprop */
```
sixstone-qq: Typo: after

python/openchange/migration/__init__.py
```diff
@@ -118,7 +118,7 @@ def unapply_version(self, app, version):
         self.db.commit()
         self._version[app] = None  # To reload the value
 
-    def migrate(self, app, target_version=None):
+    def migrate(self, app, target_version=None, extra_params=None):
```
sixstone-qq: I'd use `**kwargs` to let the subclass to define the named parameter with name meaningful instead of generic `extra_params`.

python/openchange/migration/__init__.py
```diff
@@ -148,7 +149,7 @@ def migrate(self, app, target_version=None):
             # Migrate inside a transaction
             cursor = self.db.cursor()
             if go_forward:
-                migration.apply(cursor)
+                migration.apply(cursor, extra_params)
```
sixstone-qq: Same as above

python/openchange/migration/__init__.py
```diff
@@ -267,7 +268,7 @@ class Migration(object):
     description = 'migration'
 
     @classmethod
-    def apply(cls, cur):
+    def apply(cls, cur, extra=None):
```
sixstone-qq: Same as above.

python/openchange/mailbox.py
```diff
@@ -550,3 +550,28 @@ def __init__(self, url):
...
         self._connect_to_mysql()
         self.migration_app = 'named_properties'
+
+
+class Directory(MysqlBackendMixin):
+    """
+    The directory itself does not use mysql but we have this class for ease the migrations
+    """
+
+    def __init__(self, url, setup="", ctx=""):
```
sixstone-qq: `setup` and `ctx` are not used. Could you remove it from this parameter list?

python/openchange/mailbox.py
```diff
@@ -550,3 +550,28 @@ def __init__(self, url):
...
+        self.url = url
+        self._connect_to_mysql()
+        self.migration_app = 'directory'
+
+    def migrate(self, version=None, extra=None):
```
sixstone-qq: You could use a useful name for `extra` here?

python/openchange/mailbox.py
```diff
@@ -550,3 +550,28 @@ def __init__(self, url):
...
+        """Migrate directory schema and data
+
+        :param int version: indicating which version to migrate. None migrates
+                            to the lastest version.
+        :param dict extra: in this dictionary we should have the needed data for
+                           the directory migration
```
sixstone-qq: Could you specify in this doc what is expected for this named parameter?

python/openchange/migration/__init__.py
```diff
@@ -284,4 +285,4 @@ def unapply(cls, cur):
 
 
 # Import to get the migrations from different applications
-from . import indexing, openchangedb, named_properties
+from . import indexing, openchangedb, named_properties, directory
```
sixstone-qq: Could you maintain the alphabetical order here, please?

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# OpenChangeDB DB schema and its migrations
+# Copyright (C) Enrique J. Hernández Blasco <ejhernandez@zentyal.com> 2015
```
sixstone-qq: The Copyright and the package description is incorrect. Could you copy and paste with more caution next time?

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,186 @@
...
+
+from samba.samdb import SamDB
+from samba.auth import system_session
+from samba.provision import setup_modify_ldif
+from samba.net import Net
+from samba.dcerpc import nbt
+
+# this helper functions are extracted from openchange.provision to avoid circular import
+# it would be better a refactoring of provision.ppy so we can avoid repeating code here
+
```
sixstone-qq: Moving to another package the common functions is the way to go. For instance, `openchange.utils.ldb`, I'm sure you will find a better name.

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,186 @@
...
+# end of helper functions
+
+@migration('directory', 1)
+class AddRecipientAttributes(Migration):
+    description = 'Add missing recipient attributes'
+
+    @classmethod
+    def apply(cls, cur, extra):
+        lp = extra['lp']
+        creds = extra['creds']
```
sixstone-qq: As I stated above, this is undocumented. And it must be, using `**kwargs` approach you could use it here as normal parameters like `(..., cur, lp, creds, names, setup_path)`

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,186 @@
...
+            ldif = user_upgrade_ldif_template % {'user_dn': dn}
+            user_upgrade_ldif += ldif + "\n"
+            try:
+                db.modify_ldif(ldif)
+            except Exception, ex:
+                print "Error migrating user %s: %s. Skipping user" % (dn, str(ex))
+
+    @classmethod
+    def unapply(cls, cur):
+        raise Exception("Cannot revert directory migrations")
```
sixstone-qq: Why not?

python/openchange/migration/indexing.py
```diff
@@ -30,7 +30,7 @@ class InitialIndexingMigration(Migration):
     description = 'initial'
 
     @classmethod
-    def apply(cls, cur):
+    def apply(cls, cur, extra=None):
```
sixstone-qq: Using `**kwargs` approach, svp.

python/openchange/provision.py
```diff
@@ -672,20 +693,175 @@ def accountcontrol(names, lp, creds, username=None, value=0):
...
         print "[+] Account %s enabled" % username
 
 
+def newgroup(names, lp, creds, groupname, mail=None,):
+    """extend user record with OpenChange settings.
+
+    :param lp: Loadparm context
+    :param creds: Credentials context
+    :param names: provision names object.
+    :param username: Name of user to extend
```
sixstone-qq: Incorrect documentation. You are referring to group records here...

python/openchange/provision.py
```diff
@@ -672,20 +693,175 @@ def accountcontrol(names, lp, creds, username=None, value=0):
...
+                                       "recipient_type_details": recipient_type_details,
+                                       "recipient_display_type": recipient_display_type}
+        db.modify_ldif(ldif_value)
+
+        res = db.search(base=group_dn, scope=SCOPE_BASE, attrs=["*"])
+        if len(res) == 1:
+            record = res[0]
+        else:
+            raise Exception("this should never happen as we just modified the record...")
+        record_keys = map(lambda x: x.lower(), record.keys())
```
sixstone-qq:
```python
record_keys = [x.lower() for x in record.keys()]
```
Is more pythonic.

python/openchange/provision.py
```diff
@@ -672,20 +693,175 @@ def accountcontrol(names, lp, creds, username=None, value=0):
...
+    ldif = ldif % {"group_dn": group_dn}
+    db.modify_ldif(ldif)
+
+
+def update_group(names, lp, creds, groupname):
+    """
+    Update the attributes for the group, should be called when a group is switched from security to distribution.
+    :param lp: Loadparm context
+    :param creds: Credentials context
+    :param names: provision names object.
```
sixstone-qq: Incorrect documentation for parameters...

python/openchange/provision.py
```diff
@@ -672,20 +693,175 @@ def accountcontrol(names, lp, creds, username=None, value=0):
...
+    elif security and not universal:
+        # no value for this combination, using the previous one
+        return (1024, 1)
+    elif not security and universal:
+        return (256, 1)
+    elif not security and not universal:
+        return (512, 1)
+
+
+def _smtp_user_and_domain_by_dn(dn, name, mail, names):
```
sixstone-qq: As you are returning a tuple here, I'd documentation as you have correctly done in `_group_recipient_type_details` function. It helps to readability for me and for you in 2 months.

python/openchange/provision.py
```diff
@@ -920,6 +1096,50 @@ def openchangedb_deprovision(names, lp, mapistore=None):
...
+# migration methods for openchange directory
+def directory_migrate(lp, uri=None, version=None, extra=None):
+    provisioned = False
+    try:
+        check_not_provisioned(extra['names'], lp, extra['creds'])
+    except:
+        provisioned = True
+    if not provisioned:
+        raise Exception("OpenChange is not provisioned, provision it instead of upgrading it")
+
```
sixstone-qq: In other migrations, the first migration is in charge of provision the system. You should use the same approach here and having the 1st migration as the provision one.

python/openchange/provision.py
```diff
@@ -920,6 +1096,50 @@ def openchangedb_deprovision(names, lp, mapistore=None):
     openchangedb.remove()
 
 
+# migration methods for openchange directory
+def directory_migrate(lp, uri=None, version=None, extra=None):
```
sixstone-qq: Could you use here the meaningful parameter names instead of `extra`?

setup/openchange_group
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python
+# OpenChange provision script
+#
+# Copyright (C) Jelmer Vernooij <jelmer@openchange.org> 2008
```
sixstone-qq: Incorrect copyright and package description.

setup/openchange_newuser
```diff
@@ -18,16 +18,13 @@
 #
 
 import optparse
-import os,sys
+import sys
```
sixstone-qq: :+1: to all these changes :clap:

python/openchange/migration/directory.py
javierag: A backward migration should imply remove the added attributes for elements, in some cases also remove full elements and then the schema.

It is not only that this could give a lot of work. But since the directory is  shared by others applications this could break other applications so I am not comfortable with this.

python/openchange/provision.py
javierag: No, because the provision is done with its own command and shoul not be done automatically. Moreover it seems a quite radical change.

javierag:
As for the migration I have not tested the scenario of migration-without-provision and then provision. I have made the migration resilient but this should be tested.
I will do it after doing the 'kwargs' changes

python/openchange/migration/directory.py
sixstone-qq: As a developer, I think we should provide the sysadmin the ability to perform this operation with its consequences. In order to perform the backwards migration, the sysadmin should really know what he's doing because it is a difficult path.
In addition to this, it will be as uniform as DB migrations are and they allow us to go backwards for a reason. And finally, it is the only way I see to test your code is resilient enough. Go forwards and backwards twice.

I don't see why you need to remove the full element but only the attributes added by this schema... Could you explain me the first sentence?

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,186 @@
...
+        ldb_filter = "(objectClass=user)"
+        res = db.search(base=base_dn, scope=ldb.SCOPE_SUBTREE, expression=ldb_filter)
+        for element in res:
+            dn = element.dn.get_linearized()
+            ldif = user_upgrade_ldif_template % {'user_dn': dn}
+            user_upgrade_ldif += ldif + "\n"
+            try:
+                db.modify_ldif(ldif)
+            except Exception, ex:
+                print "Error migrating user %s: %s. Skipping user" % (dn, str(ex))
```
sixstone-qq: As this is a system error, I'd use `sys.stderr` as it was designed for this purpose. (This applies to every print you put in the migration code)

javierag:
I will wait for the merge of #263 before merging this, it will conflict for sure so I will do the rebase/fixing

python/openchange/migration/directory.py
javierag: In this case, not need to remove full elements. But in future migrations you could add structural classes, for example.

javierag:
I have added migration downgrade as requested. However I get this error when removing the attributes from the schema:

  No Deleted Objects container for DN CN=ms-Exch-Recipient-Display-Type,CN=Schema,CN=Configuration,DC=kernevil,DC=lan')

I think this is really a samba issue, the downgrade and the upgrade can recover form this error.

javierag:
I have checked the delete schema error, it happens that it is not allowed to remove elements from the schema to avoid replication errors:

https://technet.microsoft.com/en-us/library/cc961741.aspx

We can mark them as defunct using the 'isDefunct' attribute but not idea if samba manages correctly this.

However I can remove 'mayContains' from a schema element and I suspect that it can trigger similar replication problems.
Maybe it was not good idea to make the downgrade after all?.

jkerihuel:
Indeed, the global logic, afair, is that an extended schema can not be cleaned up.

javierag:
Ok here we have those options:

1) Revert the downgrade option. However instead raising error this time will print a message. This is because raising error it can make other unrelated downgrades to fail
2) Remove the schema dongrade but keep the code to remove the new attributes from the user accounts. I don't see any reason to do this.
3) Use the isDefunct attribute. 

I will go for one. I don't think is worth any possible troubles that can arise when the isDefunct attribute.

jkerihuel:
@javierag :+1: with your analysis.

python/openchange/provision.py
sixstone-qq: I agree with you but I explained here in order to have a global picture. I think the scope is correct but we have to know we do not have symmetric behaviour in Database and Directory. Thus wrong assumptions may appear in the future.

sixstone-qq:
@javierag :+1: to your analysis. Could you explain it in the doc of `unapply` method somehow? It will help us in our future.

Thanks very much for taking my comments into account and the work you made to clarify this situation.

python/openchange/provision.py
```diff
@@ -621,12 +638,16 @@ def newuser(names, lp, creds, username=None, mail=None):
 add: legacyExchangeDN
 legacyExchangeDN: /o=%(firstorg)s/ou=%(firstou)s/cn=Recipients/cn=%(username)s
 add: proxyAddresses
+proxyAddresses: SMTP:%(smtp_user)s
```
sixstone-qq: Why are you changing here the order of `proxyAddresses` attr values?

javierag: In some edge cases the specification uses the proxyAddresses order to choose what of them will use.
This is not our case with NSPI, because this situation never arises but I like more this order, given priority to the smtp addresses and lowest priority to the sysadmin address.

python/openchange/provision.py
javierag: Done in line 853

python/openchange/migration/directory.py
javierag: Since those short methods and unlikely to change, I will let them her for the moment. Later, if we have another py file which use them we can move them

Makefile
javierag: I have a doubt. It will not be included in the merge when this will be merged to zentyal/openchage?

sixstone-qq: This change yes. But you have to add a change in the packaging to include this file and the packaging code is only available in `trusty` branch in Zentyal/openchange repository.

python/openchange/migration/directory.py
```diff
@@ -0,0 +1,202 @@
...
+class AddRecipientAttributes(Migration):
+    description = 'Add missing recipient attributes'
+
+    @classmethod
+    def apply(cls, cur, **kwargs):
+        lp = kwargs['lp']
+        creds = kwargs['creds']
+        names = kwargs['names']
+        setup_path = kwargs['setup_path']
+
```
javierag: from adfa import getlocalsamdb

exekias:
What is the status of this branch?

javierag:
Waiting for review

sixstone-qq:
My suggested changes have been applied and I have nothing else to say about the feature itself.

javierag:
Can we move this forward?. I am available for any question.

jkerihuel:
@javierjag: I'm reviewing the PR now

mapiproxy/servers/default/nspi/emsabp.c
```diff
@@ -633,6 +625,13 @@ _PUBLIC_ void *emsabp_query(TALLOC_CTX *mem_ctx, struct emsabp_context *emsabp_c
 		}
 		data = (void *) mvszA;
 		break;
+	case PT_LONG:
```
jkerihuel: @javierag: I don't see how this case can occur unless the ulPropTag passed to emsabp_query is just PT_LONG. I can't see any reasonably use case where this would happen. Am I missing something here?

mapiproxy/servers/default/nspi/emsabp.c
```diff
@@ -863,7 +862,9 @@ _PUBLIC_ enum MAPISTATUS emsabp_table_fetch_attrs(TALLOC_CTX *mem_ctx, struct em
 				lpProps.value.l = 0x0;
 				break;
 			case PidTagDisplayName:
-				lpProps.value.lpszW = NULL;
+				/* This value is temporal to workaround PropertyRow_addprop internals */
+				/* It will be set to NULL after the call to PropertyRow_addprop */
+				lpProps.value.lpszW = "t";
```
jkerihuel: @javierag  hummf .... really not fan of this at all.

mapiproxy/servers/default/nspi/emsabp.c
javierag: This is needed because  PR_OBJECT_TYPE and PR_DISPLAY_TYPE are not longer hardcoded and both are of this type

mapiproxy/servers/default/nspi/emsabp.c
javierag: This is because we pass a NULL to  PropertyRow_addprop the property is converted to an error property. 
The other way would we to fix PropertyRow_addprop to pass it null but this should be done as other issue. But I am not sure that the semantics of that functions allows to pass NULL values like that.

mapiproxy/servers/default/nspi/emsabp.c
jkerihuel: ohh my bad sorry, I did not see the expand icon and I thought it was attached to code around L520

mapiproxy/servers/default/nspi/emsabp.c
jkerihuel: Why don't you pass an empty string instead?

javierag: Well, 't' for temporal. But anyway is the same, just we spend one byte more (or expend 3 more if we use "temp").

jkerihuel: I disagree, this data is going to be sent back to the client, therefore can have an impact on its behavior.
What does Exchange return in this case? NULL, "", MAPI_E_NOT_FOUND?

jkerihuel: ok the "t" is indeed removed and I agree on the temporary aspect of this to be fixed after PR

javierag:
There are more issues to discuss?

jkerihuel:
:+1: for me after the minor trailing space fix in ldif file is fixed

python/openchange/provision.py
sixstone-qq: In which specification do you have this order thing? I have none.

The SMTP indicates the PRIMARY proxy address to use and the smtp ones are the SECONDARIES. Note the case. See [MS-OXCMAIL] Section 1.1 for details.

I will revert this change and provide the migration to the current status in the `directory app`. About the order, no problem on keeping it as it is if no more regressions are found.

PR #241, opened 2015-03-17
openchange_user_cleanup: clean memcached

blaxter:
It removes all keys from memcached (including no openchange values).

It could be done better (deleting only the user specific keys) but it'd need an implementation of `ccan/hash.c`. This script won't be executed frequently and memcached is a cache (O RLY?) so it doesn't hurt.

### DIFF
--- a/script/openchange_user_cleanup.py
+++ b/script/openchange_user_cleanup.py
@@ -3,6 +3,7 @@
 import imaplib
 import optparse
 import MySQLdb
+import memcache
 import ldb
 import os
 import re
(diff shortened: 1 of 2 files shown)


script/openchange_user_cleanup.py
```diff
@@ -163,6 +164,55 @@ def cleanup(self, username, password=""):
         finally:
             client.logout()
 
+class MemcachedCleaner(object):
+    def __init__(self, samba_helper=SambaOCHelper(), dry_run=False):
+        self.samba_helper = samba_helper
+        self.dry_run = dry_run
+
+    def cleanup(self, username):
+        print "===== Memcached cleanup ====="
```
sixstone-qq: I will notify the user all keys are deleted.

jkerihuel:
If the server is running while this script is executed, we would need to check whether restarting memcached  would cause some side effects or not.

script/openchange_user_cleanup.py
blaxter: done

blaxter:
I've modified indexing layer so memcached errors are ignored but logged

jkerihuel:
:hand: ignoring memcached error potentially lead on relying onto an inconsistent cache which is going to make it even harder to debug. Please revert this patch.

blaxter:
I disagree, you will have errors from memcached. For instance: cache full and a old record is deleted, **it is** a cache.

What I also want to add is to initialize memcached context with a flag that removes a server when errors happen.

jkerihuel:
We are using memcached as a just-in-time cache coupled to persistent storage (mysql). Like any other services it can fail but the only example you highlights is a bad sizing of the architecture and therefore not a justification to discard failures. 

Hence, if the problem arises when a user is synchronizing its mailbox for the first time, with your proposal we end up in message loss in the client-side copy of the user mailbox.

jkerihuel:
Moving further, our fmid cache is making use of records that are grossly around 108 bytes (uint64_t + string). If a user has a mailbox loaded with 100K items, it means we need 10.3Mb of available cache to index all this data. It would therefore mean we need 1Gb available for memcached for every 100 users.

jkerihuel:
Reading the comments again, I think we are probably not on the same page. 

What I mean in the end is that the approach makes sense as long as it is coupled to fallback mechanism to the mysql persistent storage.

exekias:
I would really consider this, you can be sure that at some point some keys may be expunged from the cache, in favor of new ones.

Even if you don't agree with that, take into account that you may update a key that it's not in the cache, for example if you just configured it and it is empty. The update will fail because the key it's not yet in the cache

jkerihuel:
@exekias: The update will not fail because it use memcached_set which create the key if not present

jkerihuel:
@exekias: Moreover with regards to expunge, we have a 30 days retention and a cache check/recreation every time the user logs in.

exekias:
What about the key delete operation?

jkerihuel:
This is a valid use case. If the cache is flushed between the moment the user logs and the user deletes the item, Outlook *should* assume the operation failed. We can eventually relax the checks on this one at emsmdb level.

sixstone-qq:
Just thinking we are using memcached here correctly. Quoting wikipedia:

    """Clients must treat Memcached as a transitory cache; they cannot assume that data stored in Memcached is still there when they need it.""

So we must fallback to a persistent solution if memcached fails or use other Key/Value store with persistency such as Couchbase Server.

PR #251, opened 2015-03-23
nspi: QueryRows now can accept a NumPos bigger or equal to table size.

javierag:
This is necessary because Outlook 2010 sometime send a NSPI QueryRow where
the initial position is after the end of the table. For example,
using the search in the GAL.
If I interpret the requested delta as going backwards, the client behavior
is fine so I suspect in a sign error in the client side.

### DIFF
--- a/mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
+++ b/mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
@@ -449,7 +449,7 @@ static void dcesrv_NspiQueryRows(struct dcesrv_call_state *dce_call,
 			goto failure;
 		}
 
-		if (r->in.pStat->Delta >= 0) {
+		if (r->in.pStat->Delta >= 0 && r->in.pStat->NumPos < ldb_res->count) {
 			start_pos = r->in.pStat->NumPos + r->in.pStat->Delta;
 			if (start_pos != 0 && start_pos >= ldb_res->count) {
 				/* inexistent position */


jkerihuel:
Have you found any reference of this behavior in MS-NSPI specifications? If so, can you reference the section in the documentation mentioning this behavior?

javierag:
No, neither in the QueryRow specifications nor in the product behavior exceptions. (the last is almost non-existent for nspi)

javierag:
This is the QueryRows input pStat parameter  with the incorrect initial position/delta:

                pStat: struct STAT
                    SortType                 : SortTypeDisplayName (0)
                    ContainerID              : 0x00000000 (0)
                    CurrentRec               : UNKNOWN_ENUM_VALUE (20495)
                    Delta                    : 9
                    NumPos                   : 0x00000016 (22)
                    TotalRecs                : 0x00000015 (21)
                    CodePage                 : 0x000004e4 (1252)
                    TemplateLocale           : 0x00000409 (1033)
                    SortLocale               : 0x00000c0a (3082)

jkerihuel:
If I read [MS-OXNSPI] specifications correctly, it states in section 2.2.8 STAT that the client can set absolute positioning value:

    If absolute positioning, as specified in section
    3.1.4.5.1, is used, the value of this field specified by the client will be ignored by the server.

It therefore more looks like 0x16 == absolute value from 0x0 to 0x15

jkerihuel:
Also, the specifications state that it is a fractional position, therefore not meant to be 100% accurate. I would therefore tend to think that the reasoning and patch is correct.

:+1: to me

javierag:
:-1: 
Thinking more about it, I think the current implementation of QueryRows is incorrect. Specification says:

   12.If the server is using the table specified by the input parameter pStat, the server MUST process
rows starting from the current position in the table specified in that parameter (including any
values of the Delta field).

It seems that this is interpreted as "trust the NumPos input parameter" but I think that we must implement absolute/relative position like in the NspiUpdateStat method. 
Maybe in this case with absolute position ot the miid 20495 the request had sense?. I think that this miid was some where around half table.

This we take a bit more of calculation in the server side but it could avoid future situtations like that.

What do you think?

jkerihuel:
@javierag: do you therefore suggest to revert this PR?

javierag:
Yes, because it only works in my case by chances, Julio has found other cases which it does not work at all.

jkerihuel:
ack'd therefore reverting the change

PR #252, opened 2015-03-24
Several fixes on copy message recipients

sixstone-qq:
* Refactor fill_prop_index to reuse as much code as possible
    * We have two new functions in libmapi to extract the email address value and display name value from a SPropTagArray checking a set of properties. This avoids duplications and is safer.
* Added checks on returned values on copying msg recipients

### DIFF
--- a/mapiproxy/servers/default/emsmdb/emsmdbp_object.c
+++ b/mapiproxy/servers/default/emsmdb/emsmdbp_object.c
@@ -875,6 +875,7 @@ static int emsmdbp_copy_properties(struct emsmdbp_context *emsmdbp_ctx, struct e
 static inline int emsmdbp_copy_message_recipients_mapistore(struct emsmdbp_context *emsmdbp_ctx, struct emsmdbp_object *source_object, struct emsmdbp_object *dest_object)
 {
 	TALLOC_CTX			*mem_ctx;
+	enum mapistore_error		ret;
 	struct mapistore_message	*msg_data;
 	uint32_t			contextID, i;
 	int				email_address_idx, display_name_idx;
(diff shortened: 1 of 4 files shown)


mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -874,26 +875,34 @@ static int emsmdbp_copy_properties(struct emsmdbp_context *emsmdbp_ctx, struct e
 
 /* FIXME: this function is already present in oxcmsg... */
 struct emsmdbp_prop_index {
-	uint32_t display_name; /* PR_DISPLAY_NAME_UNICODE or PR_7BIT_DISPLAY_NAME_UNICODE or PR_RECIPIENT_DISPLAY_NAME_UNICODE */
-	uint32_t email_address; /* PR_EMAIL_ADDRESS_UNICODE or PR_SMTP_ADDRESS_UNICODE */
+	int64_t display_name; /* PR_DISPLAY_NAME_UNICODE or PR_7BIT_DISPLAY_NAME_UNICODE or PR_RECIPIENT_DISPLAY_NAME_UNICODE */
```
jkerihuel: Why don't you simply use int instead of int64_t? The code used during assignation moreover relies on a uint32_t at the moment see LOC884

mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -4,6 +4,7 @@
    EMSMDBP: EMSMDB Provider implementation
 
    Copyright (C) Julien Kerihuel 2009-2014
+   Copyright (C) Enrique J. Hernández 2015
```
jkerihuel: Add the Copyright change to the other PR instead.

I want us to remain consistent with the way we add or not a new contributor to the file. Your other contribution definitely  requires this change. This one doesn't.

mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -918,22 +928,25 @@ static inline int emsmdbp_copy_message_recipients_mapistore(struct emsmdbp_conte
 		emsmdbp_fill_prop_index(&prop_index, msg_data->columns);
 
 		new_columns = talloc_zero(mem_ctx, struct SPropTagArray);
+		OPENCHANGE_RETVAL_IF(!new_columns, MAPI_E_NOT_ENOUGH_MEMORY, mem_ctx);
 		new_columns->cValues = msg_data->columns->cValues + 2;
 		new_columns->aulPropTag = talloc_array(new_columns, enum MAPITAGS, new_columns->cValues);
 		memcpy(new_columns->aulPropTag + 2, msg_data->columns->aulPropTag, sizeof(enum MAPITAGS) * msg_data->columns->cValues);
+		OPENCHANGE_RETVAL_IF(!new_columns->aulPropTag, MAPI_E_NOT_ENOUGH_MEMORY, mem_ctx);
```
jkerihuel: Add the OOM check directly after `talloc_array` rather than after `memcpy`

mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -918,22 +928,25 @@ static inline int emsmdbp_copy_message_recipients_mapistore(struct emsmdbp_conte
...
 		new_columns->aulPropTag[0] = PR_DISPLAY_NAME_UNICODE;
 		new_columns->aulPropTag[1] = PR_EMAIL_ADDRESS_UNICODE;
 
 		for (i = 0; i < msg_data->recipients_count; i++) {
 			new_data = talloc_array(mem_ctx, void *, new_columns->cValues);
+			OPENCHANGE_RETVAL_IF(!new_data, MAPI_E_NOT_ENOUGH_MEMORY, mem_ctx);
 			memcpy(new_data + 2, msg_data->recipients[i].data, sizeof(void *) * msg_data->columns->cValues);
-			if (prop_index.display_name != (uint32_t) -1) {
+			if (prop_index.display_name >= 0) {
```
jkerihuel: I would keep the `!= -1` instead of `>=0` which is more understandable at first glance

mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -874,26 +875,34 @@ static int emsmdbp_copy_properties(struct emsmdbp_context *emsmdbp_ctx, struct e
 
 /* FIXME: this function is already present in oxcmsg... */
 struct emsmdbp_prop_index {
-	uint32_t display_name; /* PR_DISPLAY_NAME_UNICODE or PR_7BIT_DISPLAY_NAME_UNICODE or PR_RECIPIENT_DISPLAY_NAME_UNICODE */
-	uint32_t email_address; /* PR_EMAIL_ADDRESS_UNICODE or PR_SMTP_ADDRESS_UNICODE */
+	int64_t display_name; /* PR_DISPLAY_NAME_UNICODE or PR_7BIT_DISPLAY_NAME_UNICODE or PR_RECIPIENT_DISPLAY_NAME_UNICODE */
+	int64_t email_address; /* PR_EMAIL_ADDRESS_UNICODE or PR_SMTP_ADDRESS_UNICODE */
 };
 
 static inline void emsmdbp_fill_prop_index(struct emsmdbp_prop_index *prop_index, struct SPropTagArray *properties)
```
jkerihuel: This function sounds like a useless overhead of code to me. First returning void while we in fact return data is incorrect, secondly the function does not bring much and using unit calls instead would be better and more readable to the user.

jkerihuel:
Actually, I think the code would be really improved if we just cleared all of this and replaced it with atomic calls.

sixstone-qq:
@jkerihuel ready the refactor, feel free to take a look

jkerihuel:
@sixstone-qq :+1: for the refactoring

blaxter:
:+1: nice refactoring

PR #255, opened 2015-03-26
oc_log: Add a trailing newline if one is not yet present.

jelmer:


libmapi/oc_log.c
```diff
@@ -47,7 +47,12 @@ void oc_logv(enum oc_log_level level, const char *fmt_string, va_list ap)
 		samba_level = 0;
 	}
 
-	DEBUG(samba_level, ("%s\n", line));
+	/* Add a trailing newline if one is not already present */
+	if (line[strlen(line)-1] == '\n') {
```
blaxter: `nwritten` should be `srtlen(line)`, isn't it?

jelmer: Yes. As a micro-optimization, I guess we could avoid the strlen here.

On 26 March 2015 12:32:56 pm GMT+00:00, "Jesús García Sáez" <notifications@github.com> wrote:
>> @@ -47,7 +47,12 @@ void oc_logv(enum oc_log_level level, const char
>*fmt_string, va_list ap)
>>  		samba_level = 0;
>>  	}
>>  
>> -	DEBUG(samba_level, ("%s\n", line));
>> +	/* Add a trailing newline if one is not already present */
>> +	if (line[strlen(line)-1] == '\n') {
>
>`nwritten` should be `srtlen(line)`, isn't it?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/openchange/openchange/pull/255/files#r27208283

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

blaxter: so maybe we can change it? :angel:

jelmer:
I'm not generally a fan of micro-optimizations, but sure. Updated.

blaxter:
ok, good luck reviewing pr

PR #260, opened 2015-03-29
Remove ncacn_np from OpenChange DCE/RPC endpoints.

jelmer:
Windows doesn't make these interfaces available over ncacn_np either,
and this will make it significantly easier to create a standalone openchange daemon.


jkerihuel:
:+1: Tested and works for me

PR #261, opened 2015-03-29
oc_log: Avoid call to strlen().

jelmer:


jelmer:
Actually, it seems we can't trust the value of nwritten, as nwritten is an incorrect name for this variable in  some cases.

From the manpage:
If the output was truncated due to this limit, then the  return value  is  the  number of characters (excluding the terminating null byte) which would have been written to the final string if enough space had been available.  Thus, a return value of size or more means that the output was truncated.  (See also below under NOTES.)

PR #264, opened 2015-04-08
provision: Make dns_onwner comparation case insensitive

javierag:
The provision process compares the dns_owner with the fqdn to check
if the directory is local or resides in another controller.
Since the check is with DNS names it should be case insensitive.


jkerihuel:
makes sense

javierag:
In fact Xavy found a installation where the provision failed for this problem.

sixstone-qq:
@javierag as these changes affect directly to the user, it should have an entry in `CHANGES.md` in *Fixes* section. Suggestion:

* Make provision work when DNS entries are uppercased

javierag:
This should not affect except in very edge cases. Keep in mind that the normal thing is to have the hostname in lower case, specially in our environments (zserver, zcloud).

This was detected in a zserve what was upgraded and the dns owner record somehow was uppercased.

PR #266, opened 2015-04-09
4.1 is now the stable version for the bug-analysis tool

papajulio:
Please @sixstone-qq review this it shouldn't had problems but just in case. I'm maintaining the proposed up to update just in case we start using that repo again. Additionally I added two "old" repos to maintain the differences between 4.0 and 4., It shouldn't cause problems right?


sixstone-qq:
:+1:

PR #267, opened 2015-04-09
nspi: Initialize ldb_server_sort_control correctly to avoid inverse sort

javierag:
In emsabp_ab_container_enum and emsabp_search the struct ldb_server_sort_control
was not initialized to zero.
This made that the inverse sort attribute was randomly set to non-zero, making
the address list to show with inverse order.


jkerihuel:
:+1:

PR #268, opened 2015-04-10
mapitest: zentyal_1872 expects MAPI_E_SUCCESS

blaxter:
NSPI table position functionality has been rewritten following
rules from MS-OXNSPI, in this case 3.1.4.5.1 describes absolute
positioning and NumPos value is not relevant but Delta field.

Long story short, this RoP should return MAPI_E_SUCCESS and that's
what happens now, so let's adjust the test to reality


sixstone-qq:
:+1:

PR #271, opened 2015-04-13
Fix EmptyFolder ROP

sixstone-qq:
mapistore_folder_open_folder expects a backend folder object and
not a emsmdbp object according to other calls.

This has been detected by running OXCFOLD-CREATEFOLDER mapitest.


jkerihuel:
:+1: good catch

PR #273, opened 2015-04-13
Introduce mapistore_indexing_record_add_fmid_for_uri

sixstone-qq:
This gives the possibility to set an indexing record with
a given MAPIStore URI without asking the backend to get
this URI as it is already set by OpenChange level.

This specially applies to special folders on provisioning
but also on creation of MAPIStore root folders.


jkerihuel:
this PR is going to require updates on asynchronous-notifications branch. :hand: please wait while we figure out the critical path and best way to work this around

jkerihuel:
:+1:

PR #280, opened 2015-04-15
[oxcnotif] add function to generate SOGo URL from scratch

jkerihuel:
@blaxter, @sixstone-qq, @kamenim: This should definitely be handled by the SOGo backend but this approach is the only one I can come up for now without requiring mapistore core update and associated PR in sogo code. 

Do you see anything missing/incorrect with it?

Please review :angel: :pray:


jkerihuel:
Talking with Quique, we agreed unicode and special characters such as semi-colon etc were not supported either. However unicode requires special URL encoding which definitely fits better dierctly within sogo.

The result is that I'll update the PR on best effort basis (as it's meant to be temp code) and we'll move on with the asynchronous-notifications PR.

OK with the plan?

jkerihuel:
Discussed this with @sixstone-qq. We will need this to be handled by mapistore and returned by SOGo sooner than expected. Hence we'll move on with this as a known limitation and address this properly as part of next sprint when asynchronous-notifications branch is merged into master

sixstone-qq:
:+1: as we have to address this in the backend code by modifying mapistore_interface by adding a new call... nothing else to say but in the new forthcoming PR in OpenChange and SOGo code.

PR #281, opened 2015-04-16
add old sources repository in apport-retrace conf

jacalvo:


sixstone-qq:
@jacalvo as we need only a single file, could you try to have a single apport config file for 14.04 distro?

jacalvo:
That file I've modified is the only one needed, what are you exactly proposing? Remove the current appport-config and move apport-config.backup to apport-config and also the -b option from the script?

jacalvo:
Anyway this is a must to make it work currently, and the other refactor should probably go in a different pull request...

sixstone-qq:
@jacalvo as talked in person. this refactoring can wait.

:+1:

jkerihuel:
ack'd

PR #282, opened 2015-04-16
[oxcnotif] Add asynchronous notification framework

jkerihuel:
This pull request provides the full asynchronous notification framework and removes references to all the existing and deprecated notification code. All new additions to mapistore API (session, resolver, subscription, deliver) are unit tested. The PR also includes an updated dovecot plugin for 2.2 series.


mapiproxy/dcesrv_mapiproxy.c
```diff
@@ -286,8 +287,6 @@ static void mapiproxy_op_unbind(struct dcesrv_connection_context *context, const
 		talloc_free(private);
 	}
 
-	talloc_free(context);
-
```
sixstone-qq: Sorry but I don't understand this change and how it is related with async notifications. Could you explain it to me?

jkerihuel: That is fairly understandable at it also require some context from samba4 rpc server. In a nutshell, samba rpc_server is responsible for freeing the dcesrv_connection_context, not the endpoint. 

With the patch for samba4 that addresses the termination of connection in asynchronous emsmdb, it exposes this problem more radically and causes the process to crash with an attempt to double free talloc_free data.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+	*uri = talloc_strdup(_mem_ctx, _uri);
+	MAPISTORE_RETVAL_IF(!*uri, MAPISTORE_ERR_NO_MEMORY, NULL);
+	talloc_free(_uri);
+
+	talloc_free(mem_ctx);
+	return MAPISTORE_SUCCESS;
+}
+
+/**
+   \details Retrieve properties from mapisore object
```
sixstone-qq: Typo: mapistore

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+	MAPISTORE_RETVAL_IF(!prop_data, MAPISTORE_ERR_NO_MEMORY, NULL);
+	memset(prop_data, 0, sizeof(struct mapistore_property_data) * properties->cValues);
+
+	retval = mapistore_properties_get_properties(p->mstore_ctx, contextID, context_object, mem_ctx, properties->cValues,
+						     properties->aulPropTag, prop_data);
+	MAPISTORE_RETVAL_IF(retval, retval, prop_data);
+
+	for (i = 0; i < properties->cValues; i++) {
+		if (prop_data[i].error) {
+				if (properties->aulPropTag[i] == PidTagFolderFlags) {
```
sixstone-qq: Could you document this special case? I don't know the reason of this condition.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+					*((uint32_t *)prop_data[i].data) = FolderFlags_IPM|FolderFlags_Normal;
+					data_pointers[i] = prop_data[i].data;
+					(void) talloc_reference(data_pointers, prop_data[i].data);
+					retvals[i] = MAPI_E_SUCCESS;
+				} else {
+					retvals[i] = mapistore_error_to_mapi(prop_data[i].error);
+				}
+		} else {
+			if (prop_data[i].data == NULL) {
+				if (properties->aulPropTag[i] == PidTagFolderFlags) {
```
sixstone-qq: Same as above?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+					*(uint32_t *)prop_data[i].data = FolderFlags_IPM|FolderFlags_Normal;
+					data_pointers[i] = prop_data[i].data;
+					(void) talloc_reference(data_pointers, prop_data[i].data);
+					retvals[i] = MAPI_E_SUCCESS;
+				} else {
+					retvals[i] = MAPI_E_NOT_FOUND;
+				}
+			} else {
+				if (properties->aulPropTag[i] == PidTagParentFolderId) {
+					if (*(uint64_t *)prop_data[i].data == 0x0) {
```
sixstone-qq: This is explained in get_mapistore_sogo_PidTagParentFolderId but a hint would be very positive.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+						uint64_t	parent_fid;
+
+						retval = get_mapistore_sogo_PidTagParentFolderId(mem_ctx, p, uri, &parent_fid);
+						if (retval == MAPISTORE_SUCCESS) {
+							*(uint64_t *)(prop_data[i].data) = parent_fid;
+							retvals[i] = MAPI_E_SUCCESS;
+						}
+					}
+				} else if (properties->aulPropTag[i] == PidTagFolderFlags) {
```
sixstone-qq: Same as above?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+   \param properties pointer to the array of MAPI properties to retrieve
+   \param data_pointers pointer on pointer to the data to return
+   \param retvals pointers to the return value of each data_pointers entry to return
+
+   \return MAPI_E_SUCCESS on success, otherwise MAPI error
+ */
+static enum MAPISTATUS get_properties_systemspecialfolder(TALLOC_CTX *mem_ctx, struct asyncemsmdb_private_data *p,
+							  uint64_t folderId, struct SPropTagArray *properties,
+							  void **data_pointers, enum MAPISTATUS *retvals)
+{
```
sixstone-qq: Why don't you use `get_properties_openchangedb` likewise `mapistore` is used above, instead of subfolder?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+	OC_DEBUG(0, "Notification received for session: %s", p->emsmdb_session_str);
+
+	retval = mapistore_notification_subscription_get(mem_ctx, p->mstore_ctx, p->emsmdb_uuid, &r);
+	if (retval != MAPISTORE_SUCCESS) {
+		OC_DEBUG(0, "no subscription to process");
+		talloc_free(mem_ctx);
+		return;
+	}
+
+	OC_DEBUG(0, "%d subscriptions available:", r.v.v1.count);
```
sixstone-qq: Should it be a higher number for debug level here?

mapiproxy/servers/default/emsmdb/emsmdbp.c
```diff
@@ -197,6 +197,22 @@ _PUBLIC_ struct emsmdbp_context *emsmdbp_init(struct loadparm_context *lp_ctx,
...
+   \details Associate the session uuid to the emsmdb context
+
+   \param emsmdbp_ctx pointer to the emsmdb context
+   \param uuid the uuid to associate
+
+   \return true on success, otherwise false
+ */
+_PUBLIC_ bool emsmdbp_set_session_uuid(struct emsmdbp_context *emsmdbp_ctx, struct GUID uuid)
+{
+  if (!emsmdbp_ctx) return false;
```
sixstone-qq: Wrong indentation.

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+	MAPISTORE_RETVAL_IF(!lp_ctx, MAPISTORE_ERR_INVALID_PARAMETER, NULL);
+	MAPISTORE_RETVAL_IF(!_notification_ctx, MAPISTORE_ERR_INVALID_PARAMETER, NULL);
+
+	notification_ctx = talloc_zero(mem_ctx, struct mapistore_notification_context);
+	MAPISTORE_RETVAL_IF(!notification_ctx, MAPISTORE_ERR_NO_MEMORY, NULL);
+	notification_ctx->memc_ctx = NULL;
+
+	url = lpcfg_parm_string(lp_ctx, NULL, "mapistore", "notification_cache");
+	if (url) {
+		notification_ctx->memc_ctx = memcached(url, strlen(url));
```
blaxter: A nice to have feature, due to memcached config not very straightforward we could add here [libmemcached_check_configuration](http://docs.libmemcached.org/libmemcached_configuration.html#libmemcached_check_configuration)

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: I'm sorry I don't understand what you are trying to say here. Can you rephrase please?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: In our new email / inbox refresh use case, PidTagFolderFlags is one of the properties that needs to be returned as part of the notification blob. If this value is missing, then Outlook doesn't fetch new data content. 

The hardcoded value is a reasonable choice being made to cover most/all use cases and I can't see any impact on existing code caused by this property.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: I don't think it is worth it. As you already mentioned, there is a detailed explanation about the underlying reasoning. Therefore it is just about following the code logic to figure this out, which is what we expect from any developer.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: Good point, 5 should be in use here until we define proper levels.

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: Ack'd could indeed be worth it. Will give it a look

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+	struct ndr_push				*ndr;
+	enum ndr_err_code			ndr_err_code;
+	char					*key;
+	memcached_return			rc;
+
+	/* Sanity checks */
+	MAPISTORE_RETVAL_IF(!mstore_ctx, MAPISTORE_ERR_NOT_INITIALIZED, NULL);
+	MAPISTORE_RETVAL_IF(!mstore_ctx->notification_ctx, MAPISTORE_ERR_NOT_AVAILABLE, NULL);
+	MAPISTORE_RETVAL_IF(!mstore_ctx->notification_ctx->memc_ctx, MAPISTORE_ERR_NOT_AVAILABLE, NULL);
+
```
blaxter: if cn is null ndr_* will complain about it, check missing?

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+	local_mem_ctx = talloc_new(NULL);
+	MAPISTORE_RETVAL_IF(!local_mem_ctx, MAPISTORE_ERR_NO_MEMORY, NULL);
+
+	/* Retrieve key/value */
+	retval = mapistore_notification_session_set_key(local_mem_ctx, async_uuid, &key);
+	MAPISTORE_RETVAL_IF(retval, retval, local_mem_ctx);
+
+	value = memcached_get(mstore_ctx->notification_ctx->memc_ctx, key, strlen(key), &value_len,
+			      &flags, &rc);
+	talloc_free(key);
```
blaxter: I think this is not really needed because we will free `local_mem_ctx` in any case

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+	blob.length = value_len;
+
+	ndr = ndr_pull_init_blob(&blob, local_mem_ctx);
+	MAPISTORE_RETVAL_IF(!ndr, MAPISTORE_ERR_NO_MEMORY, local_mem_ctx);
+	ndr_set_flags(&ndr->flags, LIBNDR_FLAG_NOALIGN|LIBNDR_FLAG_REF_ALLOC);
+
+	ndr_err_code = ndr_pull_mapistore_notification_session(ndr, NDR_SCALARS, &r);
+	talloc_free(ndr);
+	MAPISTORE_RETVAL_IF(ndr_err_code != NDR_ERR_SUCCESS, MAPISTORE_ERROR, local_mem_ctx);
+
```
blaxter: A check `r.vnum == 1` would be nice to be safe.

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: The documentation was lacking details and after looking at the implementation, I don't have more confidence about this call. Most of the calling functions are relying on static buffer which length is `BUFSIZ` and there are some use case where error_buffer and  error_buffer_size are not set which would require extra checks.

Too much hassle with regards to expected benefits here.

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: the unit test about session_add covers this case but I agree that a sanity check on `cn` would avoid relying on ndr_ layer. I'm trying to figure out if there was originally another reason.

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+	char					*key = NULL;
+
+	/* Sanity checks */
+	MAPISTORE_RETVAL_IF(!mstore_ctx, MAPISTORE_ERR_NOT_INITIALIZED, NULL);
+	MAPISTORE_RETVAL_IF(!cn, MAPISTORE_ERR_INVALID_PARAMETER, NULL);
+	MAPISTORE_RETVAL_IF(!host, MAPISTORE_ERR_INVALID_PARAMETER, NULL);
+	MAPISTORE_RETVAL_IF(!mstore_ctx->notification_ctx, MAPISTORE_ERR_NOT_AVAILABLE, NULL);
+	MAPISTORE_RETVAL_IF(!mstore_ctx->notification_ctx->memc_ctx, MAPISTORE_ERR_NOT_AVAILABLE, NULL);
+
+	/* memcached does not allow key with space */
```
blaxter: Would it be better to move this check in `mapistore_notification_resolver_set_key` where we are generating the key for memcached entry?

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: you're right. however afaics this pedantic free shouldn't hurt.

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: actually a switch/case would be better and anticipate addition of future versions. Will change it accordingly. This scenario may happen in the future if we do bring change but miss to update the full distributed environment resulting in conflicting versions in use.

Good catch.

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+
+	local_mem_ctx = talloc_new(NULL);
+	MAPISTORE_RETVAL_IF(!local_mem_ctx, MAPISTORE_ERR_NO_MEMORY, NULL);
+
+	retval = mapistore_notification_resolver_set_key(local_mem_ctx, cn, &key);
+	MAPISTORE_RETVAL_IF(retval, retval, local_mem_ctx);
+
+	value = memcached_get(mstore_ctx->notification_ctx->memc_ctx, key, strlen(key), &value_len,
+			      &flags, &rc);
+	talloc_free(key);
```
blaxter: I think this is not really needed because we will free `local_mem_ctx` in any case

mapiproxy/libmapistore/mapistore_notification.c
jkerihuel: makes sense

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+
+	mem_ctx = talloc_new(NULL);
+	MAPISTORE_RETVAL_IF(!mem_ctx, MAPISTORE_ERR_NO_MEMORY, NULL);
+
+	/* Prepare subscription key */
+	retval = mapistore_notification_subscription_set_key(mem_ctx, uuid, &key);
+	MAPISTORE_RETVAL_IF(retval, retval, mem_ctx);
+
+	rc = memcached_exist(mstore_ctx->notification_ctx->memc_ctx, key, strlen(key));
+	talloc_free(key);
```
blaxter: unnecessary `talloc_free`

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -3,525 +3,1355 @@
...
+
+	mem_ctx = talloc_new(NULL);
+	MAPISTORE_RETVAL_IF(!mem_ctx, MAPISTORE_ERR_NO_MEMORY, NULL);
+
+	/* Prepare deliver key */
+	retval = mapistore_notification_deliver_set_key(mem_ctx, uuid, &key);
+	MAPISTORE_RETVAL_IF(retval, retval, mem_ctx);
+
+	rc = memcached_exist(mstore_ctx->notification_ctx->memc_ctx, key, strlen(key));
+	talloc_free(key);
```
blaxter: unnecessary `talloc_free` (or unnecessary `mem_ctx`)

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -281,383 +281,17 @@ static enum MAPISTATUS dcesrv_EcDoDisconnect(struct dcesrv_call_state *dce_call,
...
-	*sizep += libmapiserver_RopNotify_size(mapi_repl);
-
-	return success;
-}
-
 static struct mapi_response *EcDoRpc_process_transaction(TALLOC_CTX *mem_ctx, 
 							 struct emsmdbp_context *emsmdbp_ctx,
-							 struct mapi_request *mapi_request)
+							 struct mapi_request *mapi_request,
+							 struct GUID uuid)
```
blaxter: `uuid` is not used but `emsmdbp_ctx->session_uuid` (and how this function is called right now they are the same)

jkerihuel: correct. good catch!

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -1906,9 +1536,58 @@ static enum MAPISTATUS dcesrv_EcDoAsyncConnectEx(struct dcesrv_call_state *dce_c
...
+	if (!handle) {
+		OC_DEBUG(0, "[EcDoAsyncConnectEx] Unable to create new handle");
+		r->out.async_handle->handle_type = 0;
+		r->out.async_handle->uuid = GUID_zero();
+		r->out.result = DCERPC_FAULT_CONTEXT_MISMATCH;
+		return MAPI_E_LOGON_FAILED;
+	}
+
+	/* Step 3. Register the global session */
+	retval = mapistore_notification_session_add(emsmdbp_ctx->mstore_ctx, r->in.handle->uuid,
```
blaxter: there is no difference, but for me it would be more clear if we use `emsmdbp_ctx->session_uuid` instead of `r->in.handle->uuid`.
And instead of `dcesrv_call_account_name(dce_call)` `emsmdbp_ctx->username`

mapiproxy/servers/default/emsmdb/oxcfold.c
```diff
@@ -140,9 +140,6 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopGetHierarchyTable(TALLOC_CTX *mem_ctx,
 	struct mapi_handles	*parent;
 	struct mapi_handles	*rec = NULL;
 	struct emsmdbp_object	*object = NULL, *parent_object = NULL;
-	struct mapistore_subscription_list *subscription_list;
-	struct mapistore_subscription *subscription;
-	struct mapistore_table_subscription_parameters subscription_parameters;
 	void			*data;
 	uint64_t		folderID;
```
blaxter: now we can remove this variable and the switch block on line 179

mapiproxy/services/plugins/dovecot/.gitignore
```diff
@@ -0,0 +1,22 @@
+.deps
+.libs
+AUTHORS
```
blaxter: This file are required for `./autogen.sh`:

    Makefile.am: error: required file './NEWS' not found
    Makefile.am: error: required file './README' not found
    Makefile.am: error: required file './AUTHORS' not found
    Makefile.am: error: required file './ChangeLog' not found
````

File: mapiproxy/libmapistore/mapistore_notification.c (hunk 11 above)
Comment (jkerihuel) [inline mapiproxy/libmapistore/mapistore_notification.c]: The original reason was to test what would be the ndr_ layer behavior if a malicious user was forging a notification with empty cn. I wanted to guarantee an error would be triggered instead of a crash - which is the current behavior.

File: mapiproxy/libmapistore/mapistore_notification.c (hunk 12 above)
Comment (jkerihuel) [inline mapiproxy/libmapistore/mapistore_notification.c]: I prefer in this function to keep the talloc_free in this function to prevent from continuously allocating memory during the execution of the function and only flush at once in the end.

Maybe too pedantic, but as said does not really hurt.

File: mapiproxy/libmapistore/mapistore_notification.c (hunk 10 above)
Comment (blaxter) [inline mapiproxy/libmapistore/mapistore_notification.c]: :+1:

File: mapiproxy/services/plugins/dovecot/.gitignore (hunk 21 above)
Comment (jkerihuel) [inline mapiproxy/services/plugins/dovecot/.gitignore]: done, thx!

File: mapiproxy/servers/default/emsmdb/oxcfold.c (hunk 20 above)
Comment (jkerihuel) [inline mapiproxy/servers/default/emsmdb/oxcfold.c]: done! thx

File: mapiproxy/dcesrv_mapiproxy.c (hunk 1 above)
Comment (sixstone-qq) [inline mapiproxy/dcesrv_mapiproxy.c]: Thanks very much for your explanation!

File: mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c (hunk 3 above)
Comment (sixstone-qq) [inline mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c]: Thanks, it'd be great to document this info in the code as a comment, but not a blocker.

File: mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c (hunk 7 above)
Comment (sixstone-qq) [inline mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c]: In `emsmdbp_object` package the naming scheme *openchangedb* and *mapistore* is used. However, here you've decided to use systemfolder which has impacted me a little bit as I see awkward with regards to coherency with the rest of the code.

Comment (jkerihuel) [inline mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c]: Ah ok. Well tbh, I have just reused the nomenclature we had in emsmdbp_object which also use systemspecialfolder. In the future , if asyncemsmdb code is growing and embeds emsmdb logic already available in emsmdb server code, I'd like to create a common layer/library/file for this.

Comment (sixstone-qq) [inline mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c]: It may be better to be discussed f2f as I don't see any procedure called systemfolder in that package...

Comment (jkerihuel) [inline mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c]: Hum emsmdbp_object_get_properties_systemspecialfolder in mapiproxy/servers/default/emsmdb/emsmdbp_object.c?

File: mapiproxy/libmapistore/mapistore_notification.c (hunk 13 above)
Comment (jkerihuel) [inline mapiproxy/libmapistore/mapistore_notification.c]: done!

File: mapiproxy/services/plugins/dovecot/openchange-plugin.c [hunk 22]
```diff
@@ -0,0 +1,373 @@
...
+
+	/* Retrieve server instances */
+	retval = mapistore_notification_resolver_get(mem_ctx, &mstore_ctx, user->username,
+						     &count, &hosts);
+	if (retval) {
+		i_debug("resolver record: %s", mapistore_errstr(retval));
+		goto end;
+	}
+
+	/* Create socket with super powers */
```
blaxter: DC or Marvel?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
sixstone-qq: You're right I'm blind, sorry for the noise :angel:

mapiproxy/services/plugins/dovecot/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+  * __openchange_cn = STRING__ This option specifies the field used by
+    the dovecot service to query the resolver service. Acceptable
+    values are: __username__ or __email__. If no option is specified,
+    username is used by default.
+
+  * __openchange_backend = STRING__ This option specifies the
+    mapistore backend used by OpenChange server to deliver and fetch
+    emails from this dovecot service. If no option is specified,
+    "sogo" is used by default.
```
sixstone-qq: I was reviewing this option as I don't understand it because, if I'm not wrong, the mapistore backend is configured in server side and nothing should be set at client side as I think dovecot plugin is.

If I'm not wrong, the reason to include it is `dcesrv_asyncemsmdb.c` Line 690. And as agreed this would be implemented with a new mapistore call in the close future.

I'm nuts or should I?

jkerihuel: Dovecot has no knowledge it is used behind sogo. Therefore if dovecot is sending a notification back to dcesrv_asyncemsmdb, we need a way to figure which folders it applies to. It would therefore require to guess/search through the list of folders. This sounded not worth the effort given that it could be solved with a simple configuration option.

The new mapistore call would generate the URI in the correct backend, which is a bit different.

mapiproxy/services/plugins/dovecot/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+plugin {
+       openchange_resolver = "--SERVER=127.0.0.1:11211"
+       openchange_cn = "username"
+}
+```
+
+  * __openchange_resolver = STRING__ This option specifies the
+    connection string used by the dovecot server to access the
+    mapisore notification resolver service. This string is currently
```
sixstone-qq: Typo: mapistore

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+	struct mapistore_context		*mstore_ctx = NULL;
+	struct GUID				uuid;
+	char					*cn = NULL;
+	const char				*ip_addr = NULL;
+	int					nn_retval;
+	size_t					sz = 0;
+	char					*bind_addr = NULL;
+	int					port = 0;
+
+	OC_DEBUG(0, "exchange_asyncemsmdb: EcDoAsyncWaitEx (0x0)");
```
sixstone-qq: Sorry for being picky, but we should use the same level (3) that we use in `dcesrv_exchange_emsmdb.c`.

jkerihuel: np, fair enough

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+}
+
+
+static NTSTATUS dcerpc_server_asyncemsmdb_bind(struct dcesrv_call_state *dce_call, const struct dcesrv_interface *iface)
+{
+	if (!openchangedb_ctx) {
+		openchangedb_ctx = mapiproxy_server_openchangedb_init(dce_call->conn->dce_ctx->lp_ctx);
+		if (!openchangedb_ctx) {
+			OC_DEBUG(OC_LOG_FATAL, "Unable to initialize openchangedb");
+			abort();
```
blaxter: I'd prefer to replicate `dcesrv_exchange_emsmdb_init` behavior and use OC_PANIC(). Basically to not have abort() calls in the code but in fault_util.c

jkerihuel: it's a valid statement. Will check if it's doable without side effects.

sixstone-qq:
:+1: I've reviewed more or less the whole code. Things I find not clear enough are spotted but none of them are blocker.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: dcesrv_asyncemsmdb links against libmapiproxy, it should therefore be straight forward to replace. change in progress

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+		if (s->v.v1.subscription[i].flags & sub_TableModified) {
+			if ((s->v.v1.subscription[i].flags & sub_WholeStore) ||
+			    (s->v.v1.subscription[i].fid == folderId)) {
+				memset(&reply, 0, sizeof(struct EcDoRpc_MAPI_REPL));
+				reply.opnum = op_MAPI_Notify;
+				reply.error_code = MAPI_E_SUCCESS;
+				reply.u.mapi_Notify.NotificationHandle = s->v.v1.subscription[i].handle;
+				reply.u.mapi_Notify.LogonId = 0;
+				reply.u.mapi_Notify.NotificationType = fnevMbit | fnevSbit | fnevTableModified;
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.TableEvent = 0x5; /* TABLE_ROW_MODIFED */
```
blaxter: Why not use `TABLE_ROW_MODIFIED` directly?, it is defined (and already available through includes `dcesrv_asyncemsmdb.h` > `mapiproxy/dcesrv_mapiproxy.h` > `gen_ndr/exchange.h`).

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.FID = folderId;
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.MID = 0x0;
```
blaxter: These `= 0x0` are not necessary (previous memset) but I guess they are there intentionally for future iterations.

jkerihuel: correct, there are also helpers to review notification code and ensure contents is conforming with expected payload to be returned. Some of them are therefore more cosmetic than anything else, while some others may have an impact in future versions.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: :+1: there was a typo

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.Instance = 0x0;
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.InsertAfterFID = 0x0;
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.InsertAfterMID = 0x0;
+				reply.u.mapi_Notify.NotificationData.ContentsTableChange.ContentsTableChangeUnion.ContentsRowModifiedNotification.InsertAfterInstance = 0x0;
+
+				SPropTagArray.cValues = s->v.v1.subscription[i].count;
+				SPropTagArray.aulPropTag = (enum MAPITAGS *)s->v.v1.subscription[i].properties;
+
+				data_pointers = talloc_array(mem_ctx, void *, SPropTagArray.cValues);
```
blaxter: There is already a method called `talloc_zero_array` that I would use instead of talloc_array+memset (it's the same as you are doing here, but more clear one call that two, isn't it?)

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1230 @@
...
+								       &mapistoreURI, true);
+				if (retval == MAPI_E_SUCCESS) {
+					retval = get_properties_systemspecialfolder(mem_ctx, p, folderId, &SPropTagArray,
+										    data_pointers, retvals);
+				} else {
+					retval = MAPI_E_SUCCESS;
+					ret = get_properties_mapistore(mem_ctx, p, folderId, &SPropTagArray,
+								       data_pointers, retvals);
+					if (ret != MAPISTORE_SUCCESS) {
+						OC_DEBUG(0, "Unable to set mapistore properties: %s", mapistore_errstr(ret));
```
blaxter: Should it be: Unable to **g**et?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+
+	return NT_STATUS_OK;
+}
+
+static NTSTATUS dcerpc_server_asyncemsmdb_unbind(struct dcesrv_connection_context *context, const struct dcesrv_interface *iface)
+{
+	enum mapistore_error			retval;
+	struct exchange_asyncemsmdb_session	*session = (struct exchange_asyncemsmdb_session *) context->private_data;
+
+	OC_DEBUG(0, "dcerpc_server_asyncemsmdb_unbind");
```
sixstone-qq: Same as above (3) debug level.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+
+	if (!session) {
+		return NT_STATUS_OK;
+	}
+
+	OC_DEBUG(0, "[asyncemsmdb] unbind %s", session->bind_addr);
```
sixstone-qq: I'll use 5 here as debug level having the previous statement as tracker or the other way around if you want.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+	retval = mapistore_notification_resolver_delete(session->mstore_ctx, session->cn, session->bind_addr);
+	if (retval != MAPISTORE_SUCCESS) {
+		OC_DEBUG(0, "[asyncemsmdb] unable to delete resolver entry %s from record %s", session->bind_addr, session->cn);
+	}
+
```
sixstone-qq: As talked in person, I'll add the `mapistore_notification_session_delete` call here even though we are using a cache as stateful machine to avoid this problem in the future *and* to give coherency to the code.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+		retval = mapistore_notification_session_get(dce_call, mstore_ctx, r->in.async_handle->uuid, &uuid, &cn);
+		if (retval != MAPISTORE_SUCCESS) {
+			OC_DEBUG(0, "[asyncemsmdb]: unable to fetch emsmdb session data");
+			*r->out.pulFlagsOut = 0x1;
+			talloc_free(mstore_ctx);
+			return NT_STATUS_OK;
+		}
+
+		/* we're allowed to reply async */
+		p = talloc(dce_call->event_ctx, struct asyncemsmdb_private_data);
```
blaxter: imho `talloc` should never be used in favor of `talloc_zero`

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+					    p);
+		if (p->fd_event == NULL) {
+			OC_DEBUG(0, "[asyncemsmdb] unable to subscribe for fd event in event loop");
+			return NT_STATUS_OK;
+		}
+	}
+
+	/* Register session  */
+	if (!session) {
+		session = talloc(asyncemsmdb_session, struct exchange_asyncemsmdb_session);
```
blaxter: same here `talloc_zero`

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+
+	/* Set the asynchronous dcerpc flag on the connection */
+	dce_call->state_flags |= DCESRV_CALL_STATE_FLAG_ASYNC;
+
+	/* Step 1. Search for an existing session */
+	session = dcesrv_find_asyncemsmdb_session(&r->in.async_handle->uuid);
+	if (session) {
+		p = (struct asyncemsmdb_private_data *) session->data;
+		/* Ensure the session is registered */
+		retval = mapistore_notification_session_exist(p->mstore_ctx, r->in.async_handle->uuid);
```
blaxter: For me it would be more easy to read `session->emsmdb_uuid`. We have already fetched our session, so better use that instead of input parameters

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
blaxter: Imho `mapistore_notification_session_delete` should be called not here but in `emsmdbp_destructor` because is there where we are calling `mapistore_notification_session_add`

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -1669,6 +1293,12 @@ static enum MAPISTATUS dcesrv_EcDoConnectEx(struct dcesrv_call_state *dce_call,
 	handle = dcesrv_handle_new(dce_call->context, EXCHANGE_HANDLE_EMSMDB);
 	OPENCHANGE_RETVAL_IF(!handle, MAPI_E_NOT_ENOUGH_RESOURCES, emsmdbp_ctx);
 
+	if (emsmdbp_set_session_uuid(emsmdbp_ctx, handle->wire_handle.uuid) == false) {
```
blaxter: Does `EcDoConnect` needs this change as well? I don't think so but I'm not sure. `EcDoConnect` will be only call in case EcDoConnectEx responds Not_Implemented or something like that? (So never?)

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: @sixstone-qq @blaxter is correct, this API is to be called from emsmdb service which is the only one responsible for creating/deleting the session.

sixstone-qq: @jkerihuel yes we have discussed this, I edited the comment :seedling:

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -0,0 +1,1229 @@
...
+	/* Set or add context */
+	retval = mapistore_search_context_by_uri(p->mstore_ctx, uri, &contextID, &context_object);
+	if (retval == MAPISTORE_SUCCESS) {
+		retval = mapistore_add_context_ref_count(p->mstore_ctx, contextID);
+	} else {
+		retval = mapistore_add_context(p->mstore_ctx, p->username, uri, folderId, &contextID, &context_object);
+	}
+	MAPISTORE_RETVAL_IF(retval, retval, NULL);
+
+	prop_data = talloc_array(NULL, struct mapistore_property_data, properties->cValues);
```
blaxter: `talloc_zero_array`?

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: not ok with this change. It is cosmetic only and imho does not improve readability of the code.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: :+1:

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
jkerihuel: If I'm not mistaken, the asynchronous interface should not be available with Outlook version using EcDoConnect/EcDoRpc. They should only be for Outlook 2000 and eventually Outlook 2003.

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
jkerihuel: @sixstone-qq @blaxter I am not comfortable with this change and would prefer to postpone this discussion after the merge. The reason is that there is still some pending discussion with Microsoft about the behavior of this function and this may impact the way we do the design in the short term.

If I were to change this now, I would have to implement a double linked list to save asyncemsmdb contexts created by EcDoAsyncConnectEx and call mapistore_notification_session_delete over all asyncemsmdb_uuid saved GUIDs. This code may potentially have to be removed after the merge which really affect its relevance imho.

I also make the assumption that the risk of collision on a random GUID() would be very small for our current and future use case and it would be acceptable for now to move forward as it is.

blaxter: works for me

sixstone-qq: As stated in my general comment of the PR, not a blocker

blaxter:
it works! :+1:

jkerihuel:
So be it! :clap: :tada: Thanks for the review guys!

PR #283, opened 2015-04-20
Ignore Memcached errors and fall back to MySQL

icandigitbaby:
* Assume Memcached keys can be missing (falling back to MySQL in that case) instead of relying on them.
* Include a cleanup of Memcached in the user cleanup script. This will flush the Memcached keys of other users that are logged in, but this won't cause OpenChange crash anymore.

The code in this PR is by @blaxter, excepting a couple minor corrections in some parentheses to fit the `OC_DEBUG` macro and a return value.


jkerihuel:
:+1:

PR #284, opened 2015-04-20
Reuse a folder as special folder if one already exists on Logon

sixstone-qq:
This is a tentative fix for Deleted Items (1) duplicated folders...

It also adds the following low level changes:
* Add `set_system_idx` call to a OpenChangeDB folder
* Fix on getting unsigned long long from a NULL string

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Unreleased changes refer to our current [master branch](https://github.com/openc
 * Sharing request and invitation of folders among different Outlook versions
 
 ### Fixes
+* No more `Deleted Items (1)`-like duplicated folders
 * Fixed creation of root folders on online mode and some special folders such as Sync Issues.
 * Fixed `Too many connections to ldap` when openchange runs on samba as member of a domain.
 * Fixed `Mark All as Read` feature regression bug introduced by 7737bdf6
--- a/mapiproxy/libmapiproxy/backends/openchangedb_ldb.c
+++ b/mapiproxy/libmapiproxy/backends/openchangedb_ldb.c
@@ -356,7 +356,7 @@ static enum MAPISTATUS set_mapistoreURI(struct openchangedb_context *self,
 	int			ret;
 	struct ldb_context	*ldb_ctx = self->data;
 
-	mem_ctx = talloc_named(NULL, 0, "get_mapistoreURI");
+	mem_ctx = talloc_named(NULL, 0, "set_mapistoreURI");
 
 	ret = ldb_search(ldb_ctx, mem_ctx, &res, ldb_get_default_basedn(ldb_ctx),
 			 LDB_SCOPE_SUBTREE, attrs, "(PidTagFolderId=%"PRIu64")", fid);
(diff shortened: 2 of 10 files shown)


mapiproxy/libmapiproxy/backends/openchangedb_ldb.c
```diff
@@ -356,7 +356,7 @@ static enum MAPISTATUS set_mapistoreURI(struct openchangedb_context *self,
 	int			ret;
 	struct ldb_context	*ldb_ctx = self->data;
 
-	mem_ctx = talloc_named(NULL, 0, "get_mapistoreURI");
+	mem_ctx = talloc_named(NULL, 0, "set_mapistoreURI");
```
blaxter: actually `talloc_new` would be the way to go. But ok

mapiproxy/servers/default/emsmdb/emsmdbp_provisioning.c
```diff
@@ -552,15 +550,25 @@ FolderId: 0x67ca828f02000001      Display Name: "                        ";  Con
...
+			} else if (ret == MAPI_E_SUCCESS) {
+				current_fid = found_fid;
+				openchangedb_set_mapistoreURI(emsmdbp_ctx->oc_ctx, username, current_fid, mapistore_url);
+				openchangedb_set_system_idx(emsmdbp_ctx->oc_ctx, username, current_fid, i);
+
+				/* change number is updated by set_folder_properties */
+			} else {
+				OC_DEBUG(1, "Error trying to get fid by name (%s): %s", current_name, mapi_get_errstr(ret));
+				/* Should we halt the process? */
+				continue;
```
blaxter: I'd put `goto error;`. Probably this will never happen but if it does there is a mayor error so no need to move forward in this  function

mapiproxy/servers/default/emsmdb/emsmdbp_provisioning.c
```diff
@@ -552,15 +550,25 @@ FolderId: 0x67ca828f02000001      Display Name: "                        ";  Con
...
-			base_name = current_name;
-			j = 1;
-			while (openchangedb_get_fid_by_name(emsmdbp_ctx->oc_ctx, username, ipm_fid, current_name, &found_fid) == MAPI_E_SUCCESS) {
-				current_name = talloc_asprintf(mem_ctx, "%s (%d)", base_name, j);
-				j++;
+			/* According to [MS-OXOSFLD] Section 3.1.4.1, the special folder must be reused
+			   if one already exists by that name */
+			ret = openchangedb_get_fid_by_name(emsmdbp_ctx->oc_ctx, username, ipm_fid, current_name, &found_fid);
+			if (ret == MAPI_E_NOT_FOUND) {
+				mapistore_indexing_get_new_folderID_as_user(emsmdbp_ctx->mstore_ctx, username, &current_fid);
```
blaxter: I know that there are lot of calls in this function that are not checking return value for errors, but if we make some change I'd prefer to do it. Either `continue` or `goto error` (I'd prefer the latter)

mapiproxy/libmapiproxy/backends/openchangedb_ldb.c
sixstone-qq: we should change the whole file for this.

mapiproxy/servers/default/emsmdb/emsmdbp_provisioning.c
sixstone-qq: I do as well, but I was conservative.

sixstone-qq:
@blaxter ready the check return values

jkerihuel:
@sixstone-qq can you rebase?

sixstone-qq:
@jkerihuel done

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Unreleased changes refer to our current [master branch](https://github.com/openc
 * Sharing request and invitation of folders among different Outlook versions
 
 ### Fixes
+* Attempt to fix `Deleted Items (1)` Duplicated folders
```
jkerihuel: I'm not ok with the wording. Either it fixes a problem or it does not, but an entry in the ChangeLog does not just "try". 

@sixstone-qq Do you think you can come up with something different?

If not, I would just remove it from the changes.md

sixstone-qq: You were right. Trying with new wording.

blaxter:
:+1:

PR #288, opened 2015-04-23
Allow test suite verbosity to be set with the `CK_VERBOSITY` environment...

javierag:
... variable

With this change we can modify the test suite verbosity like this `CK_VERBOSITY="verbose" bin/openchange-testsuite`

### DIFF
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -7,12 +7,12 @@
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


blaxter:
I don't see this change relevant for `CHANGES.md` file. Also the commit message is, again, too long. You can say the same with, e.g.: Test suite use CK_VERBOSITY env (and then link to the doc, or copypaste it http://check.sourceforge.net/doc/check_html/check_3.html#SRunner-Output). Basically you should always avoid passive voice and use imperative.

javierag:
Commit message changed as per Blaxter suggestion.

blaxter:
:+1:

sixstone-qq:
:clap:

PR #291, opened 2015-04-24
Upgrade libmemcached requirement to 1.0.18 to avoid memcached_exist issue

jkerihuel:
Avoid linking problem with 1.0.8 on Ubuntu trusty


sixstone-qq:
:+1:

PR #292, opened 2015-04-24
[mapistore] mapistore_indexing_allocate_fid does not longer fails for us...

javierag:
...ers without open connection to the indexing database.

In the case the users is not already added, instead of failing it opens a connection on behalf of the user.

### DIFF
--- a/mapiproxy/libmapistore/mapistore_indexing.c
+++ b/mapiproxy/libmapistore/mapistore_indexing.c
@@ -11,12 +11,12 @@
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


jkerihuel:
:-1: I prefer to keep the approach as it was before the change.

Do you have any information about a bug this patch would fix or what are the use case it would address?

javierag:
Yes, check issue 8591 in projects. This error was detected in a crash report.

blaxter:
This is not a solution, this is a workaround masking a problem. Either we change the way indexing is used in every call or we don't change it at all (i'd prefer the latter). The problem (the crash report which is https://tracker.zentyal.org/issues/3349) is probably because mstore context has been released and then it has an invalid indexing context.

ps: again, a commit message **too** big

javierag:
Yes, I have already discussed it  with Julien and we will not merge it.

PR #293, opened 2015-04-25
Makefile: depend on .o files rather than .c files.

jelmer:
This also requires updating some #includes that relied on the linker
line having certain -I flags.


blaxter:
Could you rebase it? When is green in travis :+1: for me

jelmer:
Unfortunately this doesn't seem quite as simple. I'm going to give up on it for now. :-(

PR #298, opened 2015-04-29
[oxcnotif] make resolver cn case insensitive

jkerihuel:
Fix case where case in Active Directory is different from the one querying the resolver service

### DIFF
--- a/mapiproxy/libmapistore/mapistore_notification.c
+++ b/mapiproxy/libmapistore/mapistore_notification.c
@@ -19,7 +19,7 @@
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-
+#include <ctype.h>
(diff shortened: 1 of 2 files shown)


sixstone-qq:
:+1:

jkerihuel:
@papajulio works for you too?

jacalvo:
I can confirm the fix, :+1: to merge this.

javierag:
Maybe it would be better to use towlower instead of tolower?  towlower is the version of tolower for wide characters.

jkerihuel:
we are only dealing with username and email address which are all utf-8 encoded.

Therefore, it is unlikely we hit a use case where utf-16 chars show up.

javierag:
ok, perfect then

PR #300, opened 2015-04-29
Add paragraph to explain how to see which unit tests are run

sixstone-qq:
Following 4e87318 commit which allow this to be set.


javierag:
Maybe it will  be better a link to http://check.sourceforge.net/doc/check_html/check_3.html#SRunner-Output so the user can see the available options?

sixstone-qq:
Done.

PR #302, opened 2015-04-30
Remove test_asyncnotif

papajulio:
As I didn't hear from it while the new notifications where developed I guess that can be removed. I found it while checking how to add to the packaging the ocnotify binary.


jkerihuel:
:-1: the test is still valid. It targets openchange client libraries. Probably require a bit of refactor/improvements but still valid.

PR #304, opened 2015-05-06
Drop kernel information from version header and mapitest.

jelmer:
Including this information prevents openchange's build from being
reproducible.

Build-time kernel information is also of limited value,
as the kernel OpenChange is built on is likely different from the
kernel it is run on.

See also https://wiki.debian.org/ReproducibleBuilds/About


blaxter:
:+1:

jkerihuel:
:+1:

PR #306, opened 2015-05-07
Set postmaster proxyAddresses entry as secondary

sixstone-qq:
According to [MS-OXCMAIL] Section 1.1, the primary
proxyAddresses entry is the SMTP one and the secondaries
proxyAddresses entries are the smtp ones.

This fixes a regression introduced in #240 in OOF support in Outlook 2007 as it was taking the secondary entry as primary to set OOF. As it is fixes a regression, no CHANGES are needed as the version is not yet published

### DIFF
--- a/python/openchange/migration/directory.py
+++ b/python/openchange/migration/directory.py
@@ -3,6 +3,7 @@
 
 # OpenChangeDB migration for directory schema and contents
 # Copyright (C) Javier Amor Garcia <jamor@zentyal.com> 2015
+#               Enrique J. Hernández <ejhernandez@zentyal.com> 2015
 #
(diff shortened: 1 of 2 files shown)


exekias:
If it's not yet published why do we need a migration? Please take into account that this kind of migrations will take *hours* in large environments, and we should avoid them, is there any way to avoid this one?

jkerihuel:
@exekias: I don't think so I'm afraid
@sixstone-qq: no similar impact on outlook > 2007?

sixstone-qq:
Not detected by QA.

blaxter:
:+1:

jkerihuel:
:+1:

PR #307, opened 2015-05-10
Fix log level on OC_DEBUG wrapper

sixstone-qq:
With this change, the log level set in samba configuration
is honored in OC_DEBUG wrapper. Before this patch, if a
developer sets the log level to 5, it requires a level 6
in samba configuration to be logged.

I think this should be rollbacked once OpenChange
uses its own debugging system.

### DIFF
--- a/libmapi/oc_log.h
+++ b/libmapi/oc_log.h
@@ -56,7 +56,7 @@ enum oc_log_level {
  * This macro is a simple wrapper around oc_log() that adds the
  * source file name and line number to the message. */
 #define OC_DEBUG(priority, format, ...) \
-	oc_log (OC_LOG_DEBUG+(priority), __location__ "(%s): " format, __PRETTY_FUNCTION__, ## __VA_ARGS__)
+	oc_log (OC_LOG_INFO+(priority), __location__ "(%s): " format, __PRETTY_FUNCTION__, ## __VA_ARGS__)
 
 /* Write a log message.
  * Like in syslog, a trailing newline is *not* required. The library will add it


jkerihuel:
This breaks the log logic for OC_LOG_FATAL, OC_LOG_ERROR, OC_LOG_WARNING and OC_LOG_INFO and sounds to me like a workaround to a temporary solution :-1:

sixstone-qq:
I'd rather prefer to have the workaround at code level with a FIXME instead of forcing the users to set an upper level. But I'd notify the users though.

blaxter:
I've been thinking about this and as it's unlikely we have an implementation of our own logging system in mid term (maybe I'm wrong), I'm in favor of changing this, with a `FIXME` comment to remember to change it.

As I see this, before this (oc_log.x) we had `DEBUG(X, ....)` now we have `OC_DEBUG(X, ...)`, until we remove `DEBUG` samba macro, OC_DEBUG(X, ..) should call `DEBUG(X, ...)` and not `DEBUG(1+X, ..) as it's doing right now.

sixstone-qq:
:+1: to @blaxter

jkerihuel:
Actually you are wrong :-) We may very likely to have a fix for this in the short-term, maybe not with Samba 4.1.18, but more potentially with 4.2.x and releases from master.

blaxter:
I don't have a strong opinion, I think is better to fix this now (including a FIXME comment) but feel free to close the pr if you disagree

jkerihuel:
@blaxter, @sixstone-qq - point taken and what you suggestions make sense, however I prefer to fix this properly directly rather than introducing intermediary fixes.

This however makes me think we seriously need to raise the logging api topic for next release.

PR #308, opened 2015-05-10
Refactor provision to use current_fid when it is set

sixstone-qq:
This fixes a regression introduced by 63f2136a when
we want to reuse a folder as special folder if one
already exists on Logon.

Without this patch, the client gets messed up because
the fid set is not corresponding to the right folder,
making appear MAPI_E_NOT_FOUND opening folders and
incongruent information in oxcfxics protocol when
hierarchy was synched.

An example was the creation the Calendar, Contacts
and Tasks special folders by the client ignoring the
one we create on first Logon.


jkerihuel:
Schedule to review this PR tomorrow

jkerihuel:
@sixstone-qq at first glance OOM check missing on `talloc_asprintf` calls

will review tomorrow scope where you use it to double check the logic.

Thanks!

blaxter:
:+1:

jkerihuel:
:+1:

PR #310, opened 2015-05-14
Named properties data migration

blaxter:
New `named properties` migration (2) to insert the initial data. This was done previously with the first client connected (and bad thing happened if clients dc in the middle of that).

The input is `libmapi/conf/mapi-named-properties` with `mparse.pl` we generate a `mapistore_namedprops.py` with an array of properties to insert into the database.

It will also work for installations with full or partial data already stored in the database (aka: it'll fix broken deploys).


jkerihuel:
@blaxter I would prefer the file to be static and added to the repository rather than auto-generated, primarily for release purposes and to track changes.

blaxter:
@jkerihuel `mparse.pl` is being already used to generate lot of stuff, like `mapistore_namedprops.ldif` which was the one used until now. I don't see any issue for releasing or track changes, you can track the changes in [the source used for the generation](https://github.com/openchange/openchange/commits/master/libmapi/conf/mapi-named-properties)

jkerihuel:
@blaxter the problem is that mapi-named-properties is deprecated and we should rely on the outputs of makepropslist script that extract tags from MS-OXPROPS and actually generate code that we statically add to the repo and let us track changes among doc revisions

blaxter:
I'm ok deprecating `mparse.pl` in favor of other method but that's another issue (which can and should be handled after this). The purpose of this pr is to move the current population of data to migration framework.

Without spending too much time if I use `makepropslist.py --pdffile \[MS-OXPROPS\].pdf` I got 515 entries, nowadays we have 709 named properties initially inserted.

jkerihuel:
ack'd, so be it for now :+1:

PR #311, opened 2015-05-15
Update openchange package version to 2.3

jkerihuel:


jkerihuel:
Preparing 2.3 OpenChange release

PR #315, opened 2015-05-20
Implement decode of request/response buffers in EcDoRpcExt2 for MAPIProxy

mmcculligh:
Also update mapitest_lzxpress_validate_test_001 to use already extracted ROP request buffer instead of decoding complete RPC method with EcDoRpcExt2.


mmcculligh:
Found another compilation error.  Fixing it.

PR #316, opened 2015-05-21
Fix property pull from data blobs in parse_ModifyRecipientRow

icandigitbaby:
When the data blob had a layout, errors were being checked at the first byte instead of at the second one.


mapiproxy/servers/default/emsmdb/oxcmsg.c
```diff
@@ -793,16 +793,18 @@ static enum MAPISTATUS oxcmsg_parse_ModifyRecipientRow(TALLOC_CTX *mem_ctx, stru
 
 	data_pos = 0;
 	for (i = 0; i < prop_count; i++) {
-
 		if (recipient_row->RecipientRow.layout) {
-			data_pos++;
+			/* As specified in [MS-OXCDATA], section 2.11.5, property values in a data blob with a
+			   layout are preceded by a byte that determines whether the property is present (0x0),
+			   missing (0x1) or if instead there is a 4-byte error code (0xa) */
```
jkerihuel: Isn't the 0x1 case missing?

jkerihuel:
:+1:

blaxter:
really nice :+1:

PR #318, opened 2015-05-22
Set postmaster proxyAddresses entry as secondary in groups

javierag:
This completes changeset [e92b1a5b19] setting postmaster address
as secondary SMTP address for group accounts.

See https://github.com/openchange/openchange/pull/306 for details

### DIFF
--- a/python/openchange/migration/directory.py
+++ b/python/openchange/migration/directory.py
@@ -187,9 +187,8 @@ add: proxyAddresses
 proxyAddresses: smtp:postmaster@{mail_domain}
 """
 
-        base_dn = "CN=Users,%s" % names.domaindn
         ldb_filter = "(proxyAddresses=*)"
-        res = db.search(base=base_dn, scope=ldb.SCOPE_SUBTREE, expression=ldb_filter)
+        res = db.search(base=names.domaindn, scope=ldb.SCOPE_SUBTREE, expression=ldb_filter)
         for element in res:
             if 'proxyAddresses' in element:
                 for entry in element['proxyAddresses']:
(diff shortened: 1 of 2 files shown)


sixstone-qq:
You miss the migration script to already deployed groups.

javierag:
I am doubting if I must expand the migration for the user or if it is better to make a third different migraton

sixstone-qq:
I'd rather go for granularity, indeed this version has been already released (2.3-VULCAN) so opting for a third migration is a must. Thank very much by the way as I forgot groups...

sixstone-qq:
Mmm, the migration script I did have worked for groups as well, so :+1:

javierag:
Well, it work as long the groups are under cn=Users, in Zentyal they are under ou=Groups

javierag:
I can change migration script to check the two alternative bases, something like this:
```
        alternative_base_dn = ["CN=Users,%s" % names.domaindn, "CN=Groups,%s" % names.domaindn]
        ldb_filter = "(proxyAddresses=*)"
        for base_dn in alternative_base_dn:
            try:
                res = db.search(base=base_dn, scope=ldb.SCOPE_SUBTREE, expression=ldb_filter)
            except ldb.LdbError as ex:
                (code, _) = ex
                if code == 32:
                    # not such dn, ignoring
                    continue
            for element in res:
               [ migration procedure for element]
```

What do you think? It is worth to carry the extra code for Zentyal server?

blaxter:
why not use `base_dn = names.domaindn`?

javierag:
The only drawback is that the search will take longer because you will walk on all the tree. 
Anyway since it only is run once and migrations could be expected to take longer time, i am ok with this.
@sixstone-qq , do you agree?

sixstone-qq:
@javierag sure

javierag:
Migration updated

blaxter:
:+1:

sixstone-qq:
:+1:

PR #319, opened 2015-05-22
Open a shared calendar from address list in Outlook 2013

sixstone-qq:
By complying a little more the spec from [MS-OXNSPI].

Namely, we have followed strictly which checks are required for every operation related with a `emsabp_search` and added some easy-to-add sanity checks. Check every individual commit for details on which server processing rule has been included in which operation.

With these fixes, no more `MAPI_E_INVALID_BOOKMARK` is returned on requesting opening a shared calendar from the address list.

### DIFF
--- a/mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
+++ b/mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
@@ -802,12 +802,31 @@ static void dcesrv_NspiGetMatches(struct dcesrv_call_state *dce_call,
 		DCESRV_NSP_RETURN(r, MAPI_E_LOGON_FAILED, NULL);
 	}
 
+	/* Step 1. Sanity checks from Server Processing Rules */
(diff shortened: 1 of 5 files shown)


mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
```diff
@@ -638,6 +655,19 @@ static void dcesrv_NspiSeekEntries(struct dcesrv_call_state *dce_call, TALLOC_CT
...
 
+	/* SortTypePhoneticDisplayName is not supported, then apply
+	   [MS-OXNSPI] Section 3.1.4.1.9 Rule 9 */
+	if (r->in.pStat->SortType == SortTypePhoneticDisplayName) {
+		retval = MAPI_E_CALL_FAILED;
+		goto failure;
+	}
+
+	if ((r->in.pStat->SortType != SortTypePhoneticDisplayName)
+	    && (r->in.pStat->SortType != SortTypePhoneticDisplayName)) {
```
blaxter: I think this should be `SortTypeDisplayName`?

blaxter:
:+1:

PR #320, opened 2015-05-22
indexing: ensure DB connection on allocating fid operation

sixstone-qq:
This avoids return MAPI_E_INVALID_PARAMETER when the connection
to the database is not done for the given user.

It mimetise other procedures in the file such as
`mapistore_indexing_record_add_fmid`.


javierag:
mapistore_indexing_add does a mapsitore_indexing_search before, so the ocrrect old behaviour is preserved and the case when the indexing is missig is covered

So ok, for me :+1:

blaxter:
:+1:

PR #323, opened 2015-05-26
mapitest: Success on NspiUpdateStat with SortTypePhoneticDisplayName

sixstone-qq:
Although it can be not supported by the server, the specification
says nothing about updating STAT block with this SortType,
so we are not longer failing


blaxter:
:+1:

PR #324, opened 2015-05-27
Fix system_idx value for special folders

jkerihuel:
Set it to -1 to match hierarchy requirements and avoid duplicated
system index.

### DIFF
--- a/mapiproxy/servers/default/emsmdb/emsmdbp_provisioning.c
+++ b/mapiproxy/servers/default/emsmdb/emsmdbp_provisioning.c
@@ -662,7 +662,7 @@ FolderId: 0x67ca828f02000001      Display Name: "                        ";  Con
 						goto error;
 					}
 				}
-				ret = openchangedb_create_folder(emsmdbp_ctx->oc_ctx, username, ipm_fid, current_fid, current_cn, mapistore_url, i);
+				ret = openchangedb_create_folder(emsmdbp_ctx->oc_ctx, username, ipm_fid, current_fid, current_cn, mapistore_url, -1);
 				if (ret != MAPI_E_SUCCESS) {
 					OC_DEBUG(1, "Error creating OpenChangeDB folder: %s", mapi_get_errstr(ret));
 					goto error;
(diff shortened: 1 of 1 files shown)


blaxter:
I'm not sure this needs to be fixed. I can't find any conflict/issue. Right now `SpecialFolders` are identified by this system_idx + `parent_folder != NULL`. Meanwhile `SystemFolders` are identified by system_idx (and returned on Logon rop).

With this change the provisioning of names will break. Though this right now is not being very useful if we merge this the provisioning names stuff should be removed.

jkerihuel:
If we don't address this fix now, it will lead to side effects in the future. The reason why we observe a globally working behavior at the moment is because get_system_idx for the mysql backend is retrieving the first record only and it fortunately happens that the first record is the correct one (because we inserted the non ipm records before the calendar, contact etc.).

Therefore, if for any reason we switch the record or have a partial cleanup, this will expose this problem but we'll waste valuable time digging in.

I prefer to fix it now.

blaxter:
No, there is no fortune there (no such thing exists). The reason why the good record is being returned is because was implemented (by me) with the constraints that I mentioned before:

 * SpecialFolders: `parent_folder != NULL`
 * SystemFolders: `parent_folder = NULL`

at least for `system_idx ∈ [0..5]` (the range of special folders).

With that info the `get_SystemFolderID` function performs the sql query sorting by parent_folder_id (and null is always first) and `get_SpecialFolderID` function performs the sql adding a `parent_folder_id IS NOT NULL` condition.

tl;dr: this change fixes nothing and breaks `openchangedb_get_SpecialFolderID` (which uses system_idx to say which special folder wants).

jkerihuel:
I see your point, thanks.

However the implementation is abusing the semantics of what system_idx was in the first place. It has been designed as a unique index helping ordering the non ipm subtree folder hierarchy.

I can see why it was reused to identify special folders and it makes sense, but not respecting the uniqueness and introducing a dependency on parent_folder is messy and confusing.

This PR is wrong, therefore closing it but will create an issue to tackle this later on in the future.

PR #327, opened 2015-05-28
Migration: add new attributes only to users with account enabled

javierag:
The msExchRecipientTypeDetails and  msExchRecipientDiplayType
attributes should only be added to users with openchange enabled
otherwise we will get error later when enabling them.


blaxter:
:+1:

PR #328, opened 2015-06-03
Provision: add --ignore-already-exists option

blaxter:
To ignore ldap errors when inserting new schemas.

Disabled by default but really useful in some scenarios.

PR #329, opened 2015-06-03
Skip duplicate schemas on provision

javierag:
Since schemas cannot be removed we have to skip already existent
schemas to allow a reprovision.

### DIFF
--- a/python/openchange/provision.py
+++ b/python/openchange/provision.py
@@ -187,7 +187,7 @@ def guess_names_from_smbconf(lp, creds=None, firstorg=None, firstou=None):
     return names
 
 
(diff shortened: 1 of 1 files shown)


javierag:
This is a slight different approach than #328 

The differences are:
- the schemas are always allow duplicates because they cannot be removed
- the configuration elements are not allowed duplicates since they can be deprovisioned
- instead of ignoring all ldap exceptions this checks whether the element exists before adding it
- In openchangedb_deprovision took ldb as default uri schema when there is none. Otherwise the deprovision can fail

What do you think @blaxter ?

blaxter:
I've integrated this commit into #328 and added the option `--ignore-already-exists` in `openchange_provision`, closing this pull request

javierag:
:clap:

PR #334, opened 2015-06-11
oxcnotif: fix notification regression in zserver

icandigitbaby:
Commit https://github.com/openchange/openchange/commit/58abea1386bfd932e29f531d7991b590f6c07ac9 introduced an error in how the URLs were built, so they worked in Zinc
but not in Zentyal server


jkerihuel:
:+1:

PR #336, opened 2015-06-15
ocdb logger: Implement ReceiveFolderTable

sixstone-qq:
Avoid crashing on GetReceiveFolderTable ROP when
openchangedb_logger is set.


blaxter:
:+1:

PR #337, opened 2015-06-15
Mapitest: Skip not implemented tests in OpenChange server

sixstone-qq:
Check using a "hack" if the server is OpenChange and skip flagged tests accordingly.

Every commit belongs to a spec with the ROPs that are not implemented.

With this PR, now *\<SUITE\>-ALL* runs all tests from the suite even if it has some of them to be skipped.


blaxter:
:ok_hand:

PR #339, opened 2015-06-17
oxcfxics: Fill MetaTagCnsetSeenFAI property only with FAI messages

sixstone-qq:
And not with normal messages information in SyncGetTransferState ROP.

According to [MS-OXCFXICS] Section 2.2.1.1.3, the meta-property must
contain only information from FAI messages and the MetaTagCnsetSeen
has information from normal messages and folders depending on the type
of the synchronisation.

This ROP is used by the client to get the latest state of synchronisation
after uploading and it is used by subsequently calls to sync contents/hierarchy.

The most common use case that is fixed:

* Old mails are now synchronized after the account cleanup

Other sync issues in folders with FAI messages (INBOX) should have
less number of sync issues.

Along with the fix, I clean it up the code:

* Return error instead of aborting when it is not necessary
* Remove dead and incorrect code
* Add more checks on returned values

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Unreleased changes refer to our current [master branch](https://github.com/openc
 * Open a shared calendar from address list in Outlook 2013
 * Send event invitation mails to several attendees, mixing internal and external recipients
 * Fix folder hierarchy synchronization issues on mailbox subfolders
+* Old mails are now synchronized after account cleanup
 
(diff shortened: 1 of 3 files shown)


jkerihuel:
I have not tested the PR yet but LGTM to me afaics :+1:

blaxter:
I've tested it and it works like a charm.

PR #340, opened 2015-06-19
oxcnotif: Implement openchange_mail_save in notification plugin

icandigitbaby:
When sending a mail to several recipients, Dovecot uses the `copy`
operation for the first recipient, then it uses the `save` one (which wasn't
implemented) for the rest of the recipients.

 The main purpose of these two functions is initialising the `msg` variable,
 so a wrapper function has been created. A couple of lines from the
 `openchange_mail_copy` function have been removed after checking they
 don't make an impact neither over LMTP nor LDA.


blaxter:
I have not tested but from code perspective seems ok :+1:

sixstone-qq:
:+1:

PR #338, opened 2015-06-15
Provision all group and other provision fixes

javierag:
This PR contains:

1. Enable all groups option when provisioning organization
2. Allow to remove openchange attributes from a user
3. Users and groups can now be in any part of the LDAP tree, this removes special code for Zental server. Unfortunately the 'enable all groups' option is dependent on a specific location.

PR #341, opened 2015-06-19
provision: Added msExchOmaAdminWirelessEnable attribute to user

javierag:
This attribute is needed to manage ActiveSync ACL per user.

Since this change is transparent to end-user there is no CHANGES.md line

### DIFF
--- a/python/openchange/migration/directory.py
+++ b/python/openchange/migration/directory.py
@@ -235,3 +235,64 @@ proxyAddresses: SMTP:postmaster@{mail_domain}
                             db.modify_ldif(ldif)
                         except Exception, ex:
                             sys.stderr.write("Error unmigrating user %s: %s\n" % (dn, ex))
+
+
+@migration('directory', 3)
+class AddActiveSyncAttributes(Migration):
+    description = 'Add missing ActiveSync attributes'
+
+    @classmethod
+    def apply(cls, cur, **kwargs):
+        from openchange.provision import get_local_samdb
+        lp = kwargs['lp']
+        creds = kwargs['creds']
+        names = kwargs['names']
+        db = get_local_samdb(names, lp, creds)
+        add_activesync_template = """
+dn: {dn}
+changetype: modify
+add: msExchOmaAdminWirelessEnable
+msExchOmaAdminWirelessEnable: 0
+"""
+        recipients = cls._get_all_enabled_users(db, names.domaindn)
+        for element in recipients:
+            if 'msExchOmaAdminWirelessEnable' in element:
+                continue
+            dn = element.dn.get_linearized()
+            ldif = add_activesync_template.format(dn=dn)
+            try:
+                db.modify_ldif(ldif)
+            except Exception, ex:
+                msg = "Error migrating recipient %s: %s\n" % (dn, ex)
+                sys.stderr.write(msg)
+
+    @classmethod
+    def unapply(cls, cur, **kwargs):
+        from openchange.provision import get_local_samdb
+        lp = kwargs['lp']
+        creds = kwargs['creds']
+        names = kwargs['names']
+        db = get_local_samdb(names, lp, creds)
+        delete_activesync_template = """
+dn: {dn}
+changetype: delete
+delete: msExchOmaAdminWirelessEnable
+"""
+        recipients = cls._get_all_enabled_users(db, names.domaindn)
+        for element in recipients:
+            if not 'msExchOmaAdminWirelessEnable' in element:
+                continue
+            dn = element.dn.get_linearized()
+            ldif = delete_activesync_template.format(dn=dn)
+            try:
+                db.modify_ldif(ldif)
+            except Exception, ex:
+                msg = "Error reverting migration in recipient %s: %s\n" % (dn, ex)
+                sys.stderr.write(msg)
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


python/openchange/migration/directory.py
```diff
@@ -26,6 +26,24 @@
...
+def _get_local_samdb(**kwargs):
+        from openchange.provision import get_local_samdb
+
+        lp = kwargs['lp']
+        creds = kwargs['creds']
+        names = kwargs['names']
+        return get_local_samdb(names, lp, creds)
+
+
+def _get_all_enabled_users(db, **kwargs):
```
sixstone-qq: I'll move this function within Migration AddActiveSyncAttributes class as it is exclusive.

python/openchange/migration/directory.py
```diff
@@ -26,6 +26,24 @@
 import sys
 
 
+def _get_local_samdb(**kwargs):
```
sixstone-qq: I see no benefit of refactoring here.

python/openchange/migration/directory.py
```diff
@@ -235,3 +253,45 @@ def unapply(cls, cur, **kwargs):
...
                             sys.stderr.write("Error unmigrating user %s: %s\n" % (dn, ex))
+
+
+@migration('directory', 3)
+class AddActiveSyncAttributes(Migration):
+    description = 'Add missing ActiveSync attributes'
+
+    @classmethod
+    def apply(cls, cur, **kwargs):
+        db = _get_local_samdb(**kwargs)
```
blaxter: `db = get_local_samdb(kwargs['lp'], kwargs['creds'], kwargs['names'])`

python/openchange/migration/directory.py
blaxter: +1

javierag:
Remember to add msExchOmaAdminWirelessEnable attributes to list of attributes to delete from user when t https://github.com/openchange/openchange/pull/338  is merged

python/openchange/migration/directory.py
javierag: I wanted to leave them outside as utility functions for future provisions, if you two dont like the idea I will inline them .

python/openchange/migration/directory.py
```diff
@@ -235,3 +235,60 @@ def unapply(cls, cur, **kwargs):
...
+@migration('directory', 3)
+class AddActiveSyncAttributes(Migration):
+    description = 'Add missing ActiveSync attributes'
+
+    @classmethod
+    def apply(cls, cur, **kwargs):
+        from openchange.provision import get_local_samdb
+        lp = kwargs['lp']
+        creds = kwargs['creds']
+        names = kwargs['names']
```
sixstone-qq: The above three lines are unnecessary and IMHO it obfuscates the code, using kwargs['lp'] and such is more concise and clearer from my POV.

sixstone-qq:
@javierag if you want to merge the delete user feature, I'd rather separate it as I don't see any discussion on that topic. Once this is done, I will add that commit here, rebase your code onto it and, finally, add new code in another commit or the same one for deleting the user use case.

python/openchange/migration/directory.py
javierag: I rather like to unroll the parameters. Also this has been done in the other migrations of the same file.

javierag:
No probel, the commentary was a note for myself to add the property to the delete user attributes when this pr is merged.

python/openchange/migration/directory.py
```diff
@@ -235,3 +235,60 @@ def unapply(cls, cur, **kwargs):
...
+            except Exception, ex:
+                msg = "Error reverting migration in recipient %s: %s\n" % (dn, ex)
+                sys.stderr.write(msg)
+
+    @classmethod
+    def _get_all_enabled_users(cls, db, base_dn):
+        ldb_filter = "(&(objectClass=user)(proxyAddresses=*))"
+        return db.search(base=base_dn,
+                         scope=ldb.SCOPE_SUBTREE,
+                         expression=ldb_filter)
```
exekias: Is the migration needed? can't we understand that the missing atribute means 0?

javierag: No because SOGO's modulesConstraint code fails the check if the attribute does not exists

exekias: what do you mean with failed? it crashes or sets to NO?

javierag: The check return false

javierag:
In the newuser and newgroup ldifs I have changed the 'add' operation for attributes to a replace operation.  The reason is that I had something else adding msExOmaAdminWirelessEnable before user provision.
Thinking about it, we want a replace operation to assure the user/group has the correct attributes, just adding a existent attr would trigger error if single-valued or had complications if multi-valued.
So I changed all to replace.

sixstone-qq:
@javierag if your reasoning is correct, I will provide another PR with the 'replace' thing or, at least, another commit. This thing has nothing to do with adding msExOmaAdminWirelessEnable attribute...

javierag:
Yes, but i need the replace to wrok wth the msExOmaAdminWirelessEnable

javierag:
I have rebased to master and I have added msExchOmaAdminWirelessEnable attribute to the list of attributes to delete

blaxter:
:+1:

PR #344, opened 2015-07-02
retrace: Added asyncemsmdb component

sixstone-qq:
After adding a new server it is good to have right categorised.

There are some crashes uncategorised that can be done right with this patch.


javierag:
Seems legit. However we use 'elif' it would not be better to add all the matchs?

sixstone-qq:
@javierag it is looping over every frame as different components can apply to the same crash. In this case, `asyncemsmdb`, `nspi`, `rfr` and `mapiproxy` are at the same level and therefore they are mutually exclusive.

javierag:
ok

PR #345, opened 2015-07-02
asyncemsmdb: Fixed return value of asyncemsmdb_mapistore_destructor

javierag:
This function should return 0 in success and -1 in error.
Before this fix it was returning 1 in success and 0 in error.
However talloc code only checks for -1 so in success the behaviour
was correct. In the other hand a error return value was not
noticed by talloc.

### DIFF
--- a/mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
+++ b/mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
@@ -100,10 +100,10 @@ static int asyncemsmdb_mapistore_destructor(void *data)
 	enum mapistore_error		retval;
 
 	retval = mapistore_release(mstore_ctx);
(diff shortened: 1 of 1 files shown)


sixstone-qq:
:+1:

javierag:
:hand: Hold on with the merge. It is possible that we delete the full destructor.

jkerihuel:
The behavior you describe and the path you've provided sounds correct. 

However I don't understand what you mean by full destructor and the blocker you are referring to. Can you come and see me when you have time to discuss this?

blaxter:
I'm going to create a puil request shortly that basically removes this code.

jkerihuel:
@blaxter before doing your PR, keep in mind that the destructor and unbind code are triggered by different code paths and removing one instance shall introduce fd / memory leaks.

javierag:
ok, closed then

PR #347, opened 2015-07-03
Free mapistore context and [async]emsmdb_sessions

blaxter:
* Fixed crash in asyncemsmdb double free of mstore_ctx members because mapistore_release called twice (on unbind and as destructor).
* mapistore context destructor removed because is not needed: all their members are created inside the same memory context tree.
* Free both asyncemsmdb_session and emsmdb_session: they are now created within different memory context and freed when they are removed from our list of active sessions.


jkerihuel:
I'm not ok with the proposed changed of removing `mapistore_release` from the API.

jkerihuel:
:+1:

sixstone-qq:
:+1: :clap:

javierag:
:clap:  This crash was hurting us a lot

PR #348, opened 2015-07-07
Advanced NDR decoder for FastTransferSourceGetBuffer response buffers

jkerihuel:
- structured output for FX markers
- structured output for FX MetaTag
- colorized (ANSI color) output for FastTransferSourceGetBuffer
- print FILETIME into human readable dates
- print boolean values as 'True' and 'False' string
- ndr oriented function for dump_data which now dumps binary blob at ndr->depth
- decidated function to supply custom decoding for important properties (PidTagChangeKey, PidTagSourceKey, PidTagPredecessorChangeList)
- add exchange_debug interface and FastTransferSourceGetBuffer meta function to easily feed ndrdump with DATA_BLOB as an input
- add a convenient script (./script/fxdump.py) to turn hexadecimal output into binary blob to feed ndrdump using the exchange_debug interface

### DIFF
--- a/exchange.idl
+++ b/exchange.idl
@@ -3519,7 +3519,7 @@ System Attendant Private Interface
 
 	/*************************/
 	/* EcDoRpc Function 0x76 */
-	typedef [flag(NDR_NOALIGN)] struct {
+	typedef [noprint,flag(NDR_NOALIGN)] struct {
 		uint32			StreamDataSize;
 		uint8			StreamData[StreamDataSize];
 	} SyncUploadStateStreamContinue_req;
--- a/exchange.idl
+++ b/exchange.idl
@@ -4714,3 +4714,14 @@ System Attendant Private Interface
 {
 	void unknown_dummy();
 }
+
+[
+ uuid("8ac6ad47-9cbb-4518-8b8d-38e755bfdf01"),
+ pointer_default(unique),
+ helpstring("Debug")
+] interface exchange_debug
+{
+  [public,noprint] void FastTransferSourceGetBuffer(
+	       [in,out][flag(NDR_REMAINING)] DATA_BLOB data
+	       );
+}
--- a/exchange.idl
+++ b/exchange.idl
@@ -60,7 +60,7 @@ cpp_quote("#include <gen_ndr/ndr_misc.h>")
 		[in,out,unique,string,charset(DOS)]     uint8   **ppszUnused,
 		[in,out,unique,string,charset(DOS)]     uint8   **ppszServer
 		);
-	
+
 	/*****************/
 	/* Function 0x01 */
 	MAPISTATUS RfrGetFQDNFromLegacyDN(
@@ -257,7 +257,7 @@ System Attendant Private Interface
 		[range(0,100000)] uint32                        cValues;
 		[size_is(cValues+1), length_is(cValues)] uint32 aulPropTag[];
 	} PropertyTagArray_r;
-    
+
 	typedef struct {
 		FlatUID_r	*lpguid;
 		uint32		ulReserved;
@@ -311,7 +311,7 @@ System Attendant Private Interface
 
 	typedef struct {
 		[range(0,100000)] uint32	cValues;
-		[size_is(cValues)] uint32	*lpl; 
+		[size_is(cValues)] uint32	*lpl;
 	} LongArray_r;
 
 	typedef struct {
@@ -321,7 +321,7 @@ System Attendant Private Interface
 
 	typedef struct {
 		[range(0,100000)]uint32		cValues;
-		[size_is(cValues)] FlatUID_r	**lpguid; 
+		[size_is(cValues)] FlatUID_r	**lpguid;
 	} FlatUIDArray_r;
 
 	typedef struct {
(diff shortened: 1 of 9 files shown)


exchange.idl
```diff
@@ -4705,3 +4714,14 @@ System Attendant Private Interface
...
+
+[
+ uuid("8ac6ad47-9cbb-4518-8b8d-38e755bfdf01"),
+ pointer_default(unique),
+ helpstring("Debug")
+] interface exchange_debug
+{
+  [public,noprint] void FastTransferSourceGetBuffer(
+	       [in,out][flag(NDR_REMAINING)] DATA_BLOB data
+	       );
```
sixstone-qq: While compiling, this option shows an error:
```
exchange.idl:4726: error: FastTransferSourceGetBuffer: [out] argument 'data' is not a pointer or array, skip client functions
``` 
Could you avoid it somehow?

ndr_fasttransfer.c
```diff
@@ -0,0 +1,467 @@
...
+#define COLOR_YELLOW   "\x1b[33m"
+#define COLOR_BLUE     "\x1b[34m"
+#define COLOR_MAGENTA  "\x1b[35m"
+#define COLOR_CYAN     "\x1b[36m"
+#define COLOR_WHITE    "\x1b[37m"
+#define COLOR_BOLD     "\x1b[1m"
+#define COLOR_INVERSE  "\x1b[7m"
+#define COLOR_BOLD_OFF "\x1b[22m"
+#define COLOR_END      "\x1b[0m"
+
```
sixstone-qq: I know that colours are fancy but they provoke reading problems using standard tools like `less`. I know I can use `less -R` and such but most of the UNIX service logs are not using colours. Sysadmins use `ccze` (https://github.com/madhouse/ccze) to colourise the log lines. We can implement a module for it.

Could we have a way to avoid using colours by default? I know they can be useful for some people, so a way to know if the display is able to print it or a configuration variable or something could be really great and an option to advance users.

sixstone-qq:
This is really great and it works in most cases. Kudos! 

However, I found the FastTransferBuffer from SyncGetTransferState ROP is not properly displayed and neither the state after pushing messageChanges. See the following example (I don't know how to help you further):
```
                                TransferBuffer           : DATA_BLOB length=52
[0000] 03 00 3A 40 02 01 96 67   1C 00 00 00 06 C9 F9 EE   ..:@...g ........
[0010] 4C 72 E9 41 97 64 8F 2A   E2 5F F0 DB 04 00 00 00   Lr.A.d.* ._......
[0020] 00 52 00 93 01 5F 50 00   03 00 17 40 00 00 00 00   .R..._P. ...@....
[0030] 03 00 3B 40                                       ..;@
                            Marker                   : "IncrSyncStateBegin" (0x403A0003)
                                0x67960102               : SBinary cb=28
                                    [0000] 06 C9 F9 EE 4C 72 E9 41   97 64 8F 2A E2 5F F0 DB   ....Lr.A .d.*._..
                                    [0010] 04 00 00 00 00 52 00 93   01 5F 50 00
                                MetaTagIdsetGiven:
                                Marker                   : "IncrSyncStateEnd" (0x403B0003)
```

exchange.idl
jkerihuel: No it is not possible

jkerihuel:
The decoder indeed require some improvements and still has number of unsupported cases but I can't find any problems in the output you have supplied. There is simply no MetaIdsetGiven data in your dump.

Can you provide additional logs where you suspect incorrect decoding?

jkerihuel:
Oh ok you mean the MetatagCnsetSeen (0x67960003) Indeed that's correct. Let me check this.

sixstone-qq:
@jkerihuel yes I meant the meta properties from OXCFXICS sorry for not being so clear on this :(

jkerihuel:
Support added for remaining ICS states Metatags: MetaTagCnsetSeenFAI, MetaTagIdsetGiven and MetaTagCnsetRead

exchange.idl
```diff
@@ -2172,12 +2177,16 @@ System Attendant Private Interface
 		[switch_is(1)] GLOBCNT	GLOBCNT;
 	} ShortTermID; /* MID, FID, CN */
 
-	typedef [public,flag(NDR_NOALIGN)] struct {
-		GUID			GUID;
-		uint8			Size;
-		uint8			Data[Size];
+	typedef [public,noprint,flag(NDR_NOALIGN)] struct {
+		GUID			NameSpaceGuid;
+		[flag(NDR_REMAINING)] DATA_BLOB LocalId;
```
sixstone-qq: This change makes SOGo-OpenChange library not compile.
```
MAPIStoreMailFolder.m:559:40: error: ‘struct XID’ has no member named ‘GUID’
   guid = [NSString stringWithGUID: &xid->GUID];
 ```
An update in SOGo library is required.

jkerihuel: Hum ... did not know SOGo was using it.

sixstone-qq: Well, I already fixed but I'm modifying this section as well, so two PRs may happen in sogo-openchange library.

PR #349, opened 2015-07-08
Workaround: Revert of release sessions

blaxter:
Partial revert of https://github.com/openchange/openchange/pull/347 (what fixes the crash there was the removal of `mapistore_release` call).

There is some issue with session handling that creates crashes if we release the session.

ps: I will look into this soon.


jkerihuel:
:+1:

PR #350, opened 2015-07-12
oxcfxics: Support REPLID-based IDSET parsing

sixstone-qq:
Added unit tests.

MetaTagIdsetDeleted used this scheme, for example.

See [MS-OXCFXICS] Section 2.2.2.4.1 for details.

PR #351, opened 2015-07-12
ndr_fasttransfer: Proper decoding for MetaTagIdsetDeleted

sixstone-qq:
It is REPLID-based and not REPLGUID-based IDSET.

It requires #350 to work.


ndr_fasttransfer.c
```diff
@@ -212,7 +213,8 @@ static int ndr_parse_ics_state(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
 		NDR_CHECK(ndr_pull_SBinary(ndr_pull, NDR_SCALARS, &PtypBinary));
 		buffer.length = PtypBinary.cb;
 		buffer.data = PtypBinary.lpb;
-		idset = IDSET_parse(mem_ctx, buffer, false);
+		idbased = (element == MetaTagIdsetDeleted);
```
jkerihuel: please use explicit ternary instead

jkerihuel:
:+1: nice catch

PR #353, opened 2015-07-15
oxcfxics: fix minimum length on parsing IDSETs

sixstone-qq:
Now the REPLID-based parsing works with instances whose length is lower than 16B.

I've picked the sanity check on GLOBSET parser to have a IDSET without GLOBSET as valid parsed. It is required to have safer code while lowering the minimum length ceil.


blaxter:
:+1:

PR #354, opened 2015-07-15
Include in IdsetDeleted only ids present in IdsetGiven

sixstone-qq:
And add unit tests to `IDSET_includes_guid_glob` function.


blaxter:
:ok_hand:

PR #358, opened 2015-07-17
oxcfxics: Always include valid mid in IdsetGiven

sixstone-qq:
Without it, the messages that are valid in the server that
the client must have the latest version are not included
in the initial IdSetGiven are returned back in the meta property.

This avoids sync issues in INBOX folder, which stops syncing, after
performing several movements at once. Although the bug is not
easy reproducible by hand, using automated tests, the bug was 100%
reproducible and with this commit is gone.

No update in `CHANGES.md` done as it is a regression from a previous unreleased change (Optimisation section).

PS: 5 weeks of several engineers to fix this.

### DIFF
--- a/mapiproxy/servers/default/emsmdb/oxcfxics.c
+++ b/mapiproxy/servers/default/emsmdb/oxcfxics.c
@@ -903,6 +903,8 @@ static bool oxcfxics_push_messageChange(struct emsmdbp_context *emsmdbp_ctx, str
 						if (retvals[1] == MAPI_E_SUCCESS) {
 							message_sync_data->cns[message_sync_data->max] = *(uint64_t *) data_pointers[1];
 						} else {
+							OC_DEBUG(5, "Unable to get change number for mid: %" PRIx64,
+								 *(uint64_t *) data_pointers[0]);
 							message_sync_data->cns[message_sync_data->max] = 0;
 						}
 						message_sync_data->max++;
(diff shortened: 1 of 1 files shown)


mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -929,11 +931,21 @@ static bool oxcfxics_push_messageChange(struct emsmdbp_context *emsmdbp_ctx, str
...
+		   notify the client the mid is still valid in the server */
+		emsmdbp_replid_to_guid(emsmdbp_ctx, owner, eid & 0xffff, &replica_guid);
+		RAWIDSET_push_guid_glob(sync_data->eid_set, &replica_guid, (eid >> 16) & 0x0000ffffffffffff);
+
 		if (folder_is_mapistore && message_sync_data->cns[message_sync_data->count] != 0) {
 			cn = ((message_sync_data->cns[message_sync_data->count] >> 16) & 0x0000ffffffffffff);
 			if (IDSET_includes_guid_glob(original_cnset_seen, &sync_data->replica_guid, cn)) {
 				synccontext->skipped_objects++;
-				OC_DEBUG(5, "message changes: cn %.16"PRIx64" already present\n", cn);
+				OC_DEBUG(5, "Skip message %"PRIx64" as cn %.12"PRIx64" already present\n", eid, cn);
```
blaxter: Wrong format, missing `%` before first `PRIx64`

blaxter:
easy fix :+1:

PR #359, opened 2015-07-17
oxcfxics: pass predecessor change list to backend on MessageMove ROP

sixstone-qq:
* Cherry-pick the proper XID structure from @jkerihuel branch to decode FXBuffer.
* Implement helper function `get_SizedXidArray` to parse `PidTagPredecessorChangeList` property value in backend code

This is send back to backend code to manage the Predecessor Change List property value sent by the client to update accordingly and manage conflicts, if required.

This should be managed by OpenChange, but by now, it is leveraged to backends.


blaxter:
:+1:

PR #360, opened 2015-07-22
OXCPERM-ALL mapitest passes against OpenChange server

sixstone-qq:
By fixing a bug in client library:

* libmapi: Set PR_ENTRYID proptag on AddUserPermission oxcperm request

And making one of the tests independent from the users available in the server.


papajulio:
:+1:

PR #361, opened 2015-07-22
mapitest: Skip all tests failing in OpenChange

sixstone-qq:
Explaining in a comment the reason because they are failing.

I could have expand the reason in separate commits, but I don't think it has any added value.


papajulio:
:+1:

icandigitbaby:
:+1:

PR #364, opened 2015-07-25
indexing: Do not setup deleted fmid in memcached

sixstone-qq:
According to `mysql_record_get_fmid`, if a uri is indexed in
memcached, then the fmid is not deleted. But the `_memcached_setup`
method is populating the key/value store with deleted entries.

This fix makes sure only no soft deleted entries are in memcached
key/value store.


jkerihuel:
:+1:

PR #368, opened 2015-07-26
Use /usr/bin/env python2.7 + MySQL Query Errors

qantourisc:


qantourisc:
Going to split the pull-request.

jelmer:
Thanks for the pull request. However, we shouldn't be using a specific minor version of Python.

At the very most, we should use "/usr/bin/env python2". That is not available on all platforms though, so I'm inclined to keep things as is. Note that upstream recommends having the 'python' command refer to python2: https://www.python.org/dev/peps/pep-0394/

qantourisc:
Well not entirely sure what do with it either. The only case where this could be a problem is when python is not installed in a standard directory, or python2 is not installed, or for some reason python refers to python3. Might be best to wait until it actually proofs to be a problem ? Then again postponing bugs is not a good idea imo. And thanks for documenting the PEP, found it once.

PR #370, opened 2015-07-27
Fixes regarding execute(query,non-tuple) causing the argument to be split up like a list

qantourisc:


blaxter:
This happens with python MySQLdb >= 1.2.5 (ubuntu has 1.2.3 packages in all its versions) due to this https://github.com/farcepest/MySQLdb1/commit/87d1145c0d6ee4f5a8ecf6d5c62d2479b9cf27ea

Could you squash both commits into one please?

blaxter:
Also there is another `cursor.execute()` to fix here https://github.com/qantourisc/openchange/blob/SQLFixes/python/openchange/migration/__init__.py#L230 Could you add it also to your commit? thanks for the catch!

blaxter:
nice

qantourisc:
Afraid there are more coming. Expect more pull requests :)

PR #371, opened 2015-07-30
Check return in oxcfxics_push_messageChange

blaxter:
This check avoids a crash in the server when the Source Key can't be
retrieved, leaving the binary structure uninitialised.

This behaviour has been seen when there are MIDs in the database with a Replica ID that cannot
exist. This patch doesn't fix that bug, it just works the crash around.


jkerihuel:
:+1:

PR #378, opened 2015-08-04
Replaced mysql_affected_rows with mysql_num_rows.

qantourisc:
This is required in case for some reason the record is already there.
If an update is requested for this record the "update" will fail.
Since no changes are made. However the value is configured at the
desired value. As such the requested action has in a way actually
succeeded.


qantourisc:
More complete fix by someone else: https://github.com/openchange/openchange/pull/379

PR #379, opened 2015-08-04
Fix MySQL update check

sixstone-qq:
`mysql_affected_rows` is only reporting the affected rows and when the value has the desired value the number of affected rows is 0. We can change this behaviour by setting CLIENT_FOUND_ROWS
flag on mysql_real_connect but I'd prefer to have the MySQL connection as standard as possible and this case is not so often.

More details on the following link:

https://dev.mysql.com/doc/refman/5.5/en/mysql-affected-rows.html

This is affecting the following low level procedures:

* set_mapistore_URI
* set_system_idx
* update_fmid

### DIFF
--- a/mapiproxy/libmapiproxy/backends/openchangedb_mysql.c
+++ b/mapiproxy/libmapiproxy/backends/openchangedb_mysql.c
@@ -320,9 +320,39 @@ static enum MAPISTATUS get_mapistoreURI(TALLOC_CTX *parent_ctx,
 	return retval;
 }
 
-/* forward declaration */
 static enum MAPISTATUS get_fid(struct openchangedb_context *self,
-			       const char *mapistore_uri, uint64_t *fidp);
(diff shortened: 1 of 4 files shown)


blaxter:
I'd rather prefer an explicit test case for this (even if it's a copy-paste of the existing one). That way it will be more clear for future readers that there was a bug about this

sixstone-qq:
@blaxter Done

mapiproxy/libmapiproxy/backends/openchangedb_mysql.c
```diff
@@ -320,6 +320,10 @@ static enum MAPISTATUS get_mapistoreURI(TALLOC_CTX *parent_ctx,
 	return retval;
 }
 
+/* forward declaration */
+static enum MAPISTATUS get_fid(struct openchangedb_context *self,
```
blaxter: why not just move the whole function here?

blaxter:
:+1:

PR #380, opened 2015-08-06
Update README.md to notify developers of the requirement

jkerihuel:
to patch packaging_impl.py from apport-retrace to retrace
packages properly from any version when the crash comes
from a Zentyal package.


sixstone-qq:
:+1: :)

PR #381, opened 2015-08-06
rpcproxy: Send Ping PDU to avoid timeout disconn in OUT channel

sixstone-qq:
As specified by [MS-RPCH] Section 3.2.4.6, we must maintain the
connection alive and we can do it by sending ping messages.

We do, by now, sending 0.9 * Timeout sent to the client in CONN/A1 RTS PDU.


exekias:
:+1:

PR #382, opened 2015-08-07
fxdecoder: Decode named properties

sixstone-qq:
Concretely, the namedPropInfo element from [MS-OXCFXICS] Section 2.2.4.1
and, then, let the property value be parsed in the caller.

See the following output examples:

* 0x91C4000B (LID 0x00008229): 'True'
* 0x905A001F (Name "Content-Type"): 'message/rfc822'

### DIFF
--- a/ndr_fasttransfer.c
+++ b/ndr_fasttransfer.c
@@ -306,6 +306,49 @@ static int ndr_parse_property(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
 	return -1;
 }
 
+/* Parse namedPropInfo element which includes:
+     PytpGuid ((%x00 PytpInteger32) | (%x01 PtypString))
+   to set the name and kind of the parsed named property
+*/
+static int ndr_parse_namedproperty(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
+				   struct ndr_pull *ndr_pull, uint32_t element,
+				   char **named_prop)
+{
+	const char  *name;
+	struct GUID PytpGuid;
+	uint8_t	    ul_kind;
+	uint32_t    dispid;
+
+	if (!named_prop) {
+		return -1;
+	}
+
+	NDR_CHECK(ndr_pull_GUID(ndr_pull, NDR_SCALARS, &PytpGuid));
+
+	NDR_CHECK(ndr_pull_uint8(ndr_pull, NDR_SCALARS, &ul_kind));
+	if (ul_kind == MNID_ID) {
+		NDR_CHECK(ndr_pull_uint32(ndr_pull, NDR_SCALARS, &dispid));
+		*named_prop = talloc_asprintf(mem_ctx, "(LID 0x%08x)", dispid);
+		if (!*named_prop) {
+			return -1;
+		}
+	} else if (ul_kind == MNID_STRING) {
+		uint32_t  _flags_save_string = ndr_pull->flags;
+		ndr_set_flags(&ndr_pull->flags, LIBNDR_FLAG_STR_NULLTERM);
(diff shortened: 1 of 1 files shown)


ndr_fasttransfer.c
```diff
@@ -261,6 +261,43 @@ static int ndr_parse_property(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
...
+		return -1;
+	}
+
+	NDR_CHECK(ndr_pull_GUID(ndr_pull, NDR_SCALARS, &PytpGuid));
+
+	NDR_CHECK(ndr_pull_uint8(ndr_pull, NDR_SCALARS, &type));
+	if (type == 0) {
+		NDR_CHECK(ndr_pull_uint32(ndr_pull, NDR_SCALARS, &dispid));
+		*named_prop = talloc_asprintf_append(*named_prop, " (LID 0x%08x)", dispid);
+	} else if (type == 1) {
```
jkerihuel: I suppose type is to specify `ulkind` value. Hence, I'd rather like to see MNID_ID and MNID_STRING in place of int values.

ndr_fasttransfer.c
```diff
@@ -261,6 +261,43 @@ static int ndr_parse_property(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
...
+	if (!named_prop) {
+		return -1;
+	}
+
+	NDR_CHECK(ndr_pull_GUID(ndr_pull, NDR_SCALARS, &PytpGuid));
+
+	NDR_CHECK(ndr_pull_uint8(ndr_pull, NDR_SCALARS, &type));
+	if (type == 0) {
+		NDR_CHECK(ndr_pull_uint32(ndr_pull, NDR_SCALARS, &dispid));
+		*named_prop = talloc_asprintf_append(*named_prop, " (LID 0x%08x)", dispid);
```
jkerihuel: missing retval check

jkerihuel: It is assuemd that named_prop has been talloc'd in the first place. I would like to use a TALLOC_CTX instead which is less prone to errors.

ndr_fasttransfer.c
```diff
@@ -261,6 +261,43 @@ static int ndr_parse_property(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
...
+	} else if (type == 1) {
+		uint32_t  _flags_save_string = ndr_pull->flags;
+		ndr_set_flags(&ndr_pull->flags, LIBNDR_FLAG_STR_NULLTERM);
+		NDR_CHECK(ndr_pull_string(ndr_pull, NDR_SCALARS, &name));
+		ndr_pull->flags = _flags_save_string;
+		*named_prop = talloc_asprintf_append(*named_prop, " (Name \"%s\")", name);
```
jkerihuel: missing retval checks

ndr_fasttransfer.c
```diff
@@ -288,10 +326,10 @@ static int ndr_parse_propValue(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
```
jkerihuel: retval checks are initially missing here. They now have an impact given that propValue isnever checked before it is used in a further talloc_ call within ndr_parse_namedproperty

ndr_fasttransfer.c
sixstone-qq: I mostly get inspiration from `fxparse.c` :dancer:

sixstone-qq:
@jkerihuel Updated the PR, let me know if you think something more useful for named properties output can be got from the given examples.

blaxter:
:+1: merging

PR #384, opened 2015-08-10
Tentative fix for CR-1136.

jkerihuel:
The analysis of the message identifiers involved in every crash
tends to highlight that the mid is initially garbage pushed to
the database and the fequency of this crash is statistically very
low ]1/5000;1/500000[. This patch therefore initializes messageID
to 0 beforehand and add memory checks to prevent data corruption.

At this stage of the investigation, this is the only plausible
explanation available. If it does fix the issue, then I expect
this fix to help narrowing it down even further.


jkerihuel:
While this is a nice to have, it is not the root cause of the issue. 

I have identified the real root cause in the logic of the synchronization code and will address it in a forthcoming patch.

blaxter:
:+1:

sixstone-qq:
:+1:

PR #385, opened 2015-08-13
Set return values in emsmdbp_object_get_properties_mapistore_root

icandigitbaby:
The array of return values wasn't being updated for all the properties.
This, plus the lack of sanity checks, was leading to crashes when there
was an error retrieving a property.

### DIFF
--- a/mapiproxy/servers/default/emsmdb/emsmdbp_object.c
+++ b/mapiproxy/servers/default/emsmdb/emsmdbp_object.c
@@ -3011,6 +3011,7 @@ static int emsmdbp_object_get_properties_mapistore_root(TALLOC_CTX *mem_ctx, str
         for (i = 0; i < properties->cValues; i++) {
 		if (properties->aulPropTag[i] == PidTagFolderFlags) {
 			folder_flags = talloc_zero(data_pointers, uint32_t);
+			OPENCHANGE_RETVAL_IF(!folder_flags, MAPI_E_NOT_ENOUGH_MEMORY, NULL);
 			*folder_flags = FolderFlags_IPM|FolderFlags_Normal;
 			data_pointers[i] = folder_flags;
 		} else if (properties->aulPropTag[i] == PidTagParentFolderId) {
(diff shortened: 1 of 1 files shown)


mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -3035,28 +3036,49 @@ static int emsmdbp_object_get_properties_mapistore_root(TALLOC_CTX *mem_ctx, str
 			}
 		}
 		else if (properties->aulPropTag[i] == PR_CONTENT_COUNT) {
-                        /* a hack to avoid fetching dynamic fields from openchange.ldb */
-                        obj_count = talloc_zero(data_pointers, uint32_t);
+			/* a hack to avoid fetching dynamic fields from openchange.ldb */
```
sixstone-qq: This comment is out of date, right?

mapiproxy/servers/default/emsmdb/emsmdbp_object.c
```diff
@@ -3126,9 +3153,9 @@ static int emsmdbp_object_get_properties_mapistore_root(TALLOC_CTX *mem_ctx, str
 				properties->aulPropTag[i] = PidTagCreationTime;
 			}
 			retval = openchangedb_get_folder_property(data_pointers, emsmdbp_ctx->oc_ctx, owner, properties->aulPropTag[i], folder->folderID, data_pointers + i);
-                }
+		}
 		retvals[i] = retval;
-        }
+    }
 
 	return MAPISTORE_SUCCESS;
```
sixstone-qq: Here the indentation fails with the latest closing curly brace.

icandigitbaby:
Thanks for the review. The (indeed) outdated comment was removed and the indentation was corrected.

blaxter:
Why `mapistore_folder_get_child_count`, `mapistore_folder_get_child_count`, `emsmdbp_folder_get_folder_count`, `emsmdbp_source_key_from_fmid`, `mapistore_properties_get_properties` are returning success if the data was not found. My concern is that we are doing here the work that needs to be implemented in each function.

And actually I think only a few of them could return `NULL` data (and/or bad return code = SUCCESS and NULL data)

icandigitbaby:
I shared the same concerns, I kept those double-checks after discussing it with @sixstone-qq and deciding that being paranoid is acceptable on this part of the code. But I agree that it reduces legibility and we should be able to rely on the functions we call, so I'm OK with removing these checks.

sixstone-qq:
:+1:

blaxter:
:ok_hand:

PR #386, opened 2015-08-16
emsmdb: Initialize attachmentID in attachment objects

javierag:
Previously the attachmentID was let initialized.
This error did not trigger any known failure because the PidTagAttachNumber
was latter assigned using the position in the attachment table.

### DIFF
--- a/mapiproxy/servers/default/emsmdb/oxcmsg.c
+++ b/mapiproxy/servers/default/emsmdb/oxcmsg.c
@@ -1422,6 +1422,7 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopCreateAttach(TALLOC_CTX *mem_ctx,
 	uint32_t		handle;
 	uint32_t		contextID;
 	uint64_t		messageID;
+	uint32_t		attachmentID = 0;
 	struct mapi_handles		*rec = NULL;
 	struct mapi_handles		*attachment_rec = NULL;
 	struct emsmdbp_object		*message_object = NULL;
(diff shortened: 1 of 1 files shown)


sixstone-qq:
:+1:

mapiproxy/servers/default/emsmdb/oxcmsg.c
```diff
@@ -1485,7 +1486,9 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopCreateAttach(TALLOC_CTX *mem_ctx,
 								   messageID, message_object);
 		if (attachment_object) {
 			mretval = mapistore_message_create_attachment(emsmdbp_ctx->mstore_ctx, contextID, message_object->backend_object,
-								      attachment_object, &attachment_object->backend_object, &mapi_repl->u.mapi_CreateAttach.AttachmentID);
+								      attachment_object, &attachment_object->backend_object, &attachmentID);
+			attachment_object->object.attachment->attachmentID = attachmentID;
```
blaxter: it would make more sense to assign `attachmentID` after checking `mretval` is ok

javierag: The specification does not say nothing about the attachmentId value in case of error.

I have chosen this layout only to parallel the one in CreateMessage and to have the less change in behaviour possible. However, like the last Julien's change in CreateMessage the AttachmentID should be initialized to avoid problems. I will assign it to 0 (first valid attachment id|)

blaxter:
Either a squash or change of commit message would be nice

javierag:
:flushed:

PR #387, opened 2015-08-18
Reuse memcached connections

blaxter:
Reuse memcached connections based on configuration string, this normally means that we will have two open connections against memcached: one for indexing and another for notifications and those connections will be reused by all sessions and operations.

### DIFF
--- a/mapiproxy/libmapistore/backends/indexing_mysql.c
+++ b/mapiproxy/libmapistore/backends/indexing_mysql.c
@@ -103,7 +103,7 @@ static memcached_st *_memcached_setup(struct indexing_context *ictx,
 	}
 
 	/* Request a memcached connection */
-	memc = oc_memcached_new_connection(conn_str);
+	memc = oc_memcached_new_connection(conn_str, true);
 	if (!memc) {
 		OC_DEBUG(1, "Error trying to get memcached connection");
 		return NULL;
@@ -773,7 +773,7 @@ static int mapistore_indexing_mysql_destructor(struct indexing_context *ictx)
 	if (ictx && ictx->data) {
 		MYSQL *conn = ictx->data;
 		if (ictx->cache) {
-			oc_memcached_release_connection((memcached_st *)ictx->cache);
+			oc_memcached_release_connection((memcached_st *)ictx->cache, true);
 		}
 		if (ictx->url) {
 			OC_DEBUG(5, "Destroying indexing context `%s`\n", ictx->url);
--- a/mapiproxy/libmapistore/backends/indexing_mysql.c
+++ b/mapiproxy/libmapistore/backends/indexing_mysql.c
@@ -11,12 +11,12 @@
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
-   
+
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
@@ -33,9 +33,9 @@
 #include "../../util/ccan/hash/hash.h"
 #include "../../util/schema_migration.h"
 #include "mapiproxy/libmapiproxy/backends/openchangedb_mysql.h"
+#include "mapiproxy/util/oc_memcached.h"
 
 #include <talloc.h>
-#include <libmemcached/memcached.h>
 
 #define MYSQL(context)	((MYSQL *)context->data)
 
@@ -88,7 +88,6 @@ static memcached_st *_memcached_setup(struct indexing_context *ictx,
 	int			mret;
 	MYSQL_RES		*res;
 	uint32_t		num_rows = 0;
-	memcached_server_st	*servers = NULL;
 	memcached_st		*memc = NULL;
 	memcached_return	rc;
 	uint32_t		i;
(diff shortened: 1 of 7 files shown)


mapiproxy/util/oc_memcached.c
```diff
@@ -0,0 +1,117 @@
...
+		rc = memcached_server_push(memc, servers);
+		if (rc != MEMCACHED_SUCCESS) {
+			OC_DEBUG(0, "[memcached] Failed to add server to memcached list");
+			return NULL;
+		}
+	}
+
+	/* This entries will never be deallocated */
+	entry = talloc_zero(talloc_autofree_context(), struct memc_item);
+	entry->config_string = talloc_strdup(entry, value_string);
```
jkerihuel: missing OOM checks

mapiproxy/libmapistore/mapistore_notification.c
```diff
@@ -103,21 +103,11 @@ enum mapistore_error mapistore_notification_init(TALLOC_CTX *mem_ctx,
...
-		servers = memcached_server_list_append(servers, MSTORE_MEMC_DFLT_HOST, 0, &rc);
-		MAPISTORE_RETVAL_IF(!servers, MAPISTORE_ERR_CONTEXT_FAILED, notification_ctx);
-
-		rc = memcached_server_push(notification_ctx->memc_ctx, servers);
-		memcached_server_list_free(servers);
-		MAPISTORE_RETVAL_IF(rc != MEMCACHED_SUCCESS, MAPISTORE_ERR_CONTEXT_FAILED, notification_ctx);
+	notification_ctx->memc_ctx = oc_memcached_new_connection(url);
+	if (!notification_ctx->memc_ctx) {
+		OC_DEBUG(1, "Error trying to get memcached connection");
+		MAPISTORE_RETVAL_ERR(MAPISTORE_ERR_NOT_INITIALIZED, NULL);
```
jkerihuel: MAPISTORE_ERR_CONTEXT_FAILED expected here

mapiproxy/util/oc_memcached.h
```diff
@@ -0,0 +1,51 @@
...
+ *  functions. Internally this connections will be reused between different
+ *  calls (it will be one connection per config_string).
+ *
+ *  \param config_string Config string to initialize a memcached context,
+ *  in case of being NULL it will default to MSTORE_MEMC_DFLT_HOST. For more
+ *  info about format check online doc of libmemcached:
+ *
+ *      http://docs.libmemcached.org/libmemcached_configuration.html
+ */
+memcached_st *oc_memcached_new_connection(const char *config_string);
```
jkerihuel: doxygen content to be moved in source file rather than header to match global usage of doxygen

mapiproxy/libmapistore/backends/indexing_mysql.c
```diff
@@ -33,6 +33,7 @@
 #include "../../util/ccan/hash/hash.h"
 #include "../../util/schema_migration.h"
 #include "mapiproxy/libmapiproxy/backends/openchangedb_mysql.h"
+#include "mapiproxy/util/oc_memcached.h"
```
jkerihuel: if oc_memcached.h is included here, then remove libmemcached/memcached.h below

mapiproxy/libmapistore/backends/indexing_mysql.c
```diff
@@ -103,21 +103,13 @@ static memcached_st *_memcached_setup(struct indexing_context *ictx,
...
-		servers = memcached_server_list_append(servers, "127.0.0.1", 11211, &rc);
-		rc = memcached_server_push(memc, servers);
-		if (rc != MEMCACHED_SUCCESS) {
-			OC_DEBUG(0, "[ERR]: Unable to add server to memcached list\n");
-			return NULL;
-		}
+	/* Request a memcached connection */
+	memc = oc_memcached_new_connection(conn_str);
+	if (!memc) {
+		OC_DEBUG(1, "Error trying to get memcached connection");
```
jkerihuel: This debug message is not relevant any longer. It is superseded by the granularity of debug logs in oc_memcached_new_connection()

jkerihuel:
:hand: From what I can see in libmemcached code, there is no mechanism to support concurrency on a shared connection or to ensure data consistency when the connection is shared. The `memcached_st` structure only holds a single `memcached_result_st` to store results.

Unless I missed something, it does not sound possible to guarantee that the data received or sent has not been modified by another connection or operations with the design proposed. Hence :-1: I'm afraid :-/

I see libmemcached is offering functions to clone an existing connection. Maybe something to look into?

blaxter:
Since when there is concurrency on openchange? I don't see any problem on samba side.

The dovecot plugin will have issues though, we don't want to reuse the connections there.

mapiproxy/libmapistore/backends/indexing_mysql.c
blaxter: It is a way to know that the memcached initialization error came from indexing layer and not from other part where `oc_memcached_new_connection` was used. I'd rather keep this message and maybe increase the log level of the log messages inside `oc_memcached_new_connection`

jkerihuel:
openchange is not using the samba event loop properly, that is a fact. When we do, I see a problem and I don't want to revert/undo code properly designed in the first place.

Moving back to the initial matter of this PR, do you have any evidence that our use our memcached is causing leaks on file descriptors? How many connections does a client initiate on memcached server per session?

Is the problem you are trying to fix simply caused by a lack of available file descriptors given default limit or is the existing code causing this besides its requirements per session?

mapiproxy/libmapistore/backends/indexing_mysql.c
jkerihuel: Make sense, but the error message does not highlight it comes from the indexing layer. Wouldn't it be more accurate to pass the name of the layer to adapt underlying debug messages?

blaxter:
This issue (which is the same that we had several months ago with mysql) is because rigth now sessions are not properly freed so you can have hundreds of open connections to memcached (each user uses 3).

blaxter:
And how is openchange using badly the samba event loop? What I understand is that we fill a `struct dcesrv_interface` with our functions (like `mapiproxy_op_dispatch`) and then the dcerpc server will have its event loop and will call our functions but when an operation (e.g. any `dcesrv_EcDoRpcExt2`) is being executed there is no way a change of context can happen.

In fact if concurrency is possible, I've been always assuming it isn't, there will be bugs related with mysql usage for sure.

exekias:
I agree with @blaxter here, openchange source is full of code that doesn't take into account event loop, nor threading. Achieving any of them would require many changes.

With this PR we are fixing a known issue in terms of scalability, while keeping open a "going back" to threaded mode (with the threading flag https://github.com/openchange/openchange/pull/387/files#diff-2394642bbeb59e66bacf5fde91c22f80R267).

What do you think?

jkerihuel:
@blaxter, @exekias: You have forgotten about the asyncemsmdb endpoint which is not managed by mapiproxy, which is asynchronous, registers callback in samba tevent loop which are triggered when data arrives on a listening socket and make use of mapistore notification contexts.

If the root cause is sessions not being freed properly and the concern being scalability, then it does not make sense to keep propagating workarounds instead of investing time fixing the real cause once for all.

blaxter:
afaik we are in a single thread event loop environment, no real concurrency is happening. In the case of asyncemsmdb endpoint when `EcDoAsyncWaitEx_handler` function is being executed no other operation can happen until the end of that function, to be able to do that (several operations running at the same time) we have to use the callback mechanism of the event loop framework (tevent in this case).

blaxter:
I've integrated this in https://github.com/Zentyal/openchange repository

PR #388, opened 2015-08-21
Notifications when username is different from email

blaxter:
* Use email (obtained from `proxyAddresses:SMTP:xxxx` entry which must be always present for openchange users) to create notification session.
* Removed `openchange_cn` option from dovecot plugin, always using email.

This fixes scenarios where the username is different from the email, for example username `bob` with an email `bobby@domain.com`.

### DIFF
--- a/mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
+++ b/mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
@@ -1542,6 +1542,7 @@ static enum MAPISTATUS dcesrv_EcDoAsyncConnectEx(struct dcesrv_call_state *dce_c
 	struct dcesrv_handle		*handle;
 	struct dcesrv_endpoint		*item;
 	bool				found;
+	const char			*email;
 
 	OC_DEBUG(3, "exchange_emsmdb: EcDoAsyncConnectEx (0xe)\n");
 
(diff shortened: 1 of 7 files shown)


mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -1542,12 +1542,14 @@ static enum MAPISTATUS dcesrv_EcDoAsyncConnectEx(struct dcesrv_call_state *dce_c
...
 	struct dcesrv_endpoint		*item;
 	bool				found;
+	const char			*email;
 
 	OC_DEBUG(3, "exchange_emsmdb: EcDoAsyncConnectEx (0xe)\n");
 
 	/* Step 0. Ensure incoming user is authenticated */
 	if (!dcesrv_call_authenticated(dce_call)) {
 		OC_DEBUG(1, "No challenge requested by client, cannot authenticate");
+	failure:
```
jkerihuel: I would prefer you to revert the `failure: and goto failure` change because I don't like it and it is detrimental to readability imho

blaxter: is easier to read the same 4 lines over and over again when there is an error on the same function? For me this is the best scenario and justification to use `goto` (and is used a lot of times on openchange code, `grep -rn "goto failure"`)

jkerihuel: I find it more readable the way it was before and I am not going to have a discussion for a 12 lines of pseudo-optimization. Please revert it.

jkerihuel:
I assume it would work the same if we have virtual domains ... given that you have a working environment to test it, can you check up on that?

My only concern is about notifications that are not newmail ones. Every new notifications we introduce would have to be based on proxyAddresses instead of user cn, I am not sure it is going to work.

jkerihuel:
I have doubts about using ProxyAddresses to identify the user session uniquely. I suppose you have already got through this, so you can probably shed some light on it:

1. Why not using the mail attribute in place of ProxyAddresses SMTP entry? Are they different?

2. ProxyAddresses is an attribute with multiple values for which this patch only retrieves the first one matching the pattern. According to the MSDN documentation, it sounds like 2 at least can be set, one for inbound and the other for outbound traffic. Questions:

* do we *always* only set 1 SMTP enty for ProxyAddresses?
* can it be remotely modified by a user or administrator?
* is this value unique across the samba AD or can we have duplicates?

blaxter:
From openchange point of view mail attribute is not mandatory, proxyAddresses SMTP it is. ProxyAddresses can have multiple values but only one with "SMTP:X". We always set 1 SMTP:XX entry for ProxyAddresses, an user administrator can edit anything, and yes it is unique.

mapiproxy/services/plugins/dovecot/ChangeLog
```diff
@@ -0,0 +1 @@
+ * Removed openchange_cn option, always using email
```
sixstone-qq: As we are not using mail attribute, but proxyAddresses one, we must specify that in the changelog to whomever read this...

sixstone-qq:
:+1: apart from the minor comment.

mapiproxy/services/plugins/dovecot/ChangeLog
blaxter: I disagree, this is dovecot plugin changelog, here we don't know at all proxyAddresses attribute.

sixstone-qq: Ok, let it as it is.

blaxter:
This pull request is not very important (just an improvement) so I want to wait for Julien's feedback before merging.

blaxter:
pr moved here https://github.com/zentyal/openchange/pull/157

PR #389, opened 2015-08-21
Free user sessions on unbind

exekias:
This PR unifies session handling (under mpm_session) for nspi, emsmdb and asyncemsmdb.

After researching on RPC sessions, along with @blaxter, we decided to implement session releasing this way:

 * When a session is created, it's assigned to the server_id, context_id it was created on
 * On unbind, server_id, context_id pair is marked as released
 * When all user contexts are marked as released, underlying sessions are released too

This will ensure that we don't free a session that would be reused on a server_id, context_id different from the one it was created on (and that happens on common clients). Delaying session freeing to the last channel close.

### DIFF
--- a/mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
+++ b/mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
@@ -44,7 +44,7 @@ static struct mpm_session	*mpm_sessions = NULL;
    \return Pointer to an allocated mpm_session structure on success,
    otherwise NULL
  */
-struct mpm_session *mpm_session_new(struct server_id serverid,
+static struct mpm_session *mpm_session_new(struct server_id serverid,
 				    uint32_t context_id,
 				    const char *username,
 				    struct GUID	*uuid)
--- a/mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
+++ b/mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
@@ -10,12 +10,12 @@
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
-   
+
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
@@ -33,44 +33,6 @@
 
 static struct mpm_session	*mpm_sessions = NULL;
 
-/**
-   \details Create and return an allocated pointer to a mpm session
-
-   \param mem_ctx pointer to the memory context
-   \param serverid reference to the session context server identifier
-   structure
-   \param context_id reference to the context identifier
-
-   \return Pointer to an allocated mpm_session structure on success,
-   otherwise NULL
- */
-static struct mpm_session *mpm_session_new(struct server_id serverid,
-				    uint32_t context_id,
-				    const char *username,
-				    struct GUID	*uuid)
-{
-	struct mpm_session	*session = NULL;
-
-	session = talloc_zero(NULL, struct mpm_session);
-	if (!session) return NULL;
-
-	session->server_id = serverid;
-	session->context_id = context_id;
-	if (uuid) session->uuid = *uuid;
-	session->destructor = NULL;
-	session->private_data = NULL;
-
-	/* Released RPC connection, session may be kept to avoid premature release
-	of private_data, that will be cleared when all user sessions are released */
-	session->released = false;
-	session->username = talloc_strdup(session, username);
-	if (!session->username) return NULL;
-
-	DLIST_ADD_END(mpm_sessions, session, struct mpm_session *);
-
-	return session;
-}
-
 
 /**
    \details Create and return an allocated pointer to a mpm session
@@ -84,14 +46,51 @@ static struct mpm_session *mpm_session_new(struct server_id serverid,
 struct mpm_session *mpm_session_init(struct dcesrv_call_state *dce_call,
 				     struct GUID *uuid)
 {
+	struct mpm_session	*session = NULL;
+
 	if (!dce_call) return NULL;
 	if (!dce_call->conn) return NULL;
 	if (!dce_call->context) return NULL;
 
-	return mpm_session_new(dce_call->conn->server_id,
-			       dce_call->context->context_id,
-			       dcesrv_call_account_name(dce_call),
-			       uuid);
+	session = talloc_zero(NULL, struct mpm_session);
+	if (!session) return NULL;
+
+	session->server_id = dce_call->conn->server_id;
+	session->context_id = dce_call->context->context_id;
+	if (uuid) {
+		session->uuid = *uuid;
+	}
+	session->destructor = NULL;
+	session->private_data = NULL;
+
+	/* Released RPC connection, session may be kept to avoid premature release
+	of private_data, that will be cleared when all user sessions are released */
+	session->released = false;
+	session->username = talloc_strdup(session, dcesrv_call_account_name(dce_call));
+	if (!session->username) return NULL;
+	DLIST_ADD(mpm_sessions, session);
+
+	return session;
+}
+
+
+/**
+   \details Free session private data
+
+   \param session to free
+
+   \return true if data was released
+ */
+bool mpm_session_release(struct mpm_session *session) {
+	if (!session) return false;
+
+	if (session->destructor) {
+		session->destructor(session->private_data);
+		session->destructor = NULL;
+		return true;
+	}
+
+	return false;
 }
 
 
@@ -135,7 +134,7 @@ bool mpm_session_set_private_data(struct mpm_session *session,
 
 
 /**
-   \details Release RPC session
+   \details Unbind RPC session
 
    This code will release mapiproxy session data (and remove it from
    mpm_sessions array) when all user sessions are marked as deleted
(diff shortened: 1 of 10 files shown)


blaxter:
rebase master please

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -40,63 +44,54 @@
    \return Pointer to an allocated mpm_session structure on success,
    otherwise NULL
  */
-struct mpm_session *mpm_session_new(TALLOC_CTX *mem_ctx, 
-				     struct server_id serverid,
-				     uint32_t context_id)
+struct mpm_session *mpm_session_new(struct server_id serverid,
```
blaxter: This method should be private, it's only used at `mpm_session_init`

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -40,63 +44,54 @@
...
-	if (!mem_ctx) return NULL;
-
-	session = talloc_zero(mem_ctx, struct mpm_session);
+	session = talloc_zero(NULL, struct mpm_session);
 	if (!session) return NULL;
 
 	session->server_id = serverid;
 	session->context_id = context_id;
-	session->ref_count = 0;
+	if (uuid) session->uuid = *uuid;
```
blaxter: It'd be safer to create a copy and avoid taking the reference pointer. It's the same case as username field below

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -40,63 +44,54 @@
...
-   \details Increment the ref_count associated to a session
-
-   \param session pointer to the session where to increment ref_count
-
-   \return true on success, otherwise false
- */
-bool mpm_session_increment_ref_count(struct mpm_session *session)
-{
-	if (!session) return false;
+	// Released RPC connection, session may be kept to avoid premature release
```
blaxter: Even for inline comments we normally use `/* */`

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -140,42 +135,88 @@ bool mpm_session_set_private_data(struct mpm_session *session,
...
-		return false;
+	struct mpm_session	*current = NULL;
+	bool			release = true;
+	char			*username = NULL;
+
+	if (!server_id) return false;
+
+	// Mark all server_id/context_id sessions as released
+	for (current = mpm_sessions; current; current = current->next) {
+		if (memcmp(server_id, &current->server_id, sizeof(struct server_id)) == 0 &&
```
blaxter: It exists a `server_id_equal` function

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -140,42 +135,88 @@ bool mpm_session_set_private_data(struct mpm_session *session,
...
+			for (current = mpm_sessions; current; current = current->next) {
+				if (strcmp(username, current->username) == 0) {
+					// Destroy private data
+					if (current->destructor) {
+						current->destructor(current->private_data);
+					}
+
+					// Remove and free session entry
+					DLIST_REMOVE(mpm_sessions, current);
+					talloc_free(current);
```
blaxter: You are releasing `current` and then accessing `current->next` on the for loop, this should be change. I'd suggest using `while` loop instead

mapiproxy/libmapiproxy/dcesrv_mapiproxy_session.c
```diff
@@ -40,63 +44,54 @@
...
-{
-	if (!session) return false;
+	// Released RPC connection, session may be kept to avoid premature release
+	// of private_data, that will be cleared when all user sessions are released
+	session->released = false;
+	session->username = talloc_strdup(session, username);
+	if (!session->username) return NULL;
 
-	session->ref_count += 1;
+	DLIST_ADD_END(mpm_sessions, session, struct mpm_session *);
```
blaxter: Any reason inserting at the end? Personally I'd rather prefer `DLIST_ADD` because the `DLIST_ADD_END` macro has a third unused parameter. Not very important though.

mapiproxy/libmapiproxy/libmapiproxy.h
```diff
@@ -186,14 +190,14 @@ TDB_CONTEXT *mapiproxy_server_emsabp_tdb_init(struct loadparm_context *);
...
-bool mpm_session_increment_ref_count(struct mpm_session *);
+struct mpm_session *mpm_session_new(struct server_id, uint32_t, const char *, struct GUID *);
+struct mpm_session *mpm_session_init(struct dcesrv_call_state *, struct GUID *);
 bool mpm_session_set_destructor(struct mpm_session *, bool (*destructor)(void *));
 bool mpm_session_set_private_data(struct mpm_session *, void *);
-bool mpm_session_release(struct mpm_session *);
+bool mpm_session_unbind(struct server_id *, uint32_t);
 bool mpm_session_cmp_sub(struct mpm_session *, struct server_id, uint32_t);
 bool mpm_session_cmp(struct mpm_session *, struct dcesrv_call_state *);
+struct mpm_session *mpm_session_find_by_uuid(struct GUID *uuid);
```
blaxter: Personally I don't like it, but remove the name of the parameter like the rest of the definitions on the header files.

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -256,23 +226,6 @@ static enum MAPISTATUS dcesrv_EcDoDisconnect(struct dcesrv_call_state *dce_call,
 		return MAPI_E_LOGON_FAILED;
 	}
 
-	/* Step 1. Retrieve handle and free if emsmdbp context and session are available */
```
blaxter: In an explicit disconnect operation sent by the client I think we should still release the session as usual

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
```diff
@@ -35,34 +36,20 @@
 struct exchange_emsmdb_session		*emsmdb_session = NULL;
```
blaxter: This is not longer used

mapiproxy/servers/default/nspi/dcesrv_exchange_nsp.c
```diff
@@ -212,23 +187,6 @@ static void dcesrv_NspiUnbind(struct dcesrv_call_state *dce_call,
 		DCESRV_NSP_RETURN(r, MAPI_E_LOGON_FAILED, NULL);
 	}
 
-	/* Step 1. Retrieve handle and free if emsabp context and session are available */
```
blaxter: Still think that in an explicit disconnect from a client we should release the session (this is an operation not an unbind from rpc layer)

blaxter:
Right now, because of the `talloc_free` under `for` loop, the sessions are not been released (only one).

But what I don't like is to keep associating our sessions (identified by a random generated GUID) to `server_id`+ `context_id` (which is the way dcerpc identifies a channel) because a session can be used through a different channel that the one was created, so the info that is being stored on `struct mpm_session` (server_id and context_id) is actually in which channel the session was created (which is not very useful because of the former statement). If we keep it like this we cannot release easily sessions on explicit disconnect/unbind operations (a disconnect of a session could go through a different server_id that a mpm_session has).

imho we should keep two things to handle this properly: (1) the sessions (GUID) like now and (2) a list/hash/whatever of `[server_id, context_id, username]`.

mapiproxy/servers/default/emsmdb/dcesrv_exchange_emsmdb.c
exekias: I had my doubts here, as I didn't want to do same thing from different places, but It looks reasonable as explicit disconnects seems to happen even when the user keeps using the client

mapiproxy/modules/mpm_cache.c
```diff
@@ -386,14 +387,14 @@ static NTSTATUS cache_pull_OpenMessage(struct dcesrv_call_state *dce_call,
 
 	message = talloc((TALLOC_CTX *)mpm, struct mpm_message);
 	NT_STATUS_HAVE_NO_MEMORY(message);
-	
-	message->session = mpm_session_init((TALLOC_CTX *)mpm, dce_call);
+
+	message->session = mpm_session_init(dce_call, NULL);
 	NT_STATUS_HAVE_NO_MEMORY(message->session);
```
jkerihuel: This call and all the other mpm_session_init calls in this file are going to cause openchange to crash.

You specify a struct GUID * pointer being NULL and there is no check done anywhere before assignment at https://github.com/openchange/openchange/pull/389/files#diff-1acfc358c59edcd26a813334c163524cR59

exekias: The check is in the assignment itself:
```
if (uuid) session->uuid = *uuid;
```

Is that what you refer to?

jkerihuel: You are correct, there is indeed a check (btw, it does not respect the coding policy). 

However this is prone to errors and it can still causes crashes elsewhere such as in mpm_session_find_by_uuid: https://github.com/openchange/openchange/pull/389/files#diff-1acfc358c59edcd26a813334c163524cR203

In this function the session->uuid is not checked beforehand and will lead to a crash in GUID_equal.

If uuid is a required parameter, then it needs to be handled as such: add sanity checks on parameters, return an error if it is missing and propagate session creation failure to upper layers.

exekias: It is handled there also, as GUID is initialized to zeros, if you try to find by UUID it will not match that one (except for the case where try to find by zeroed UUID...)

exekias: Clarification to my previous comment: session->uuid is not a pointer but the UUID value

exekias: Please pass me a link to the policy, I did 2 mistakes already on this PR and I would like to avoid doing more in the future :)

blaxter: Just a note, session->uuid is `struct GUID` not `struct GUID *`

exekias: I will add a input check, if UUID parameter is NULL -> return NULL, that is something that can happen for real

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.h
```diff
@@ -4,6 +4,7 @@
    OpenChange Project
 
    Copyright (C) Julien Kerihuel 2015
+   Copyright (C) Carlos Pérez-Aradros Herce 2015
```
jkerihuel: I am uncomfortable with all these Author changes.

It is not your fault but I think OpenChange clearly needs to state what is the difference between an author and a contributor and how it reflects on the source code.

An author is a contributor to the file but a contributor is not necessary an author of the file.

From my perspective, an author should only be added to a file if the contribution is significant and contacting the author to be required or useful to understand the in-depth logic. Authoring is all about usefulness and authority for people looking at the code and who want to find answers.

 This change for example or the mpm_cache.c do not fit into this category imho. However I would definitely do a git blame and contact the contributor if I have questions about few lines of code.

mapiproxy/modules/mpm_cache.c
jkerihuel: I'd love to have a link to an established policy for OpenChange but you already know there is no such written document. However we have hundreds of thousands of lines of code and this line looks nothing similar to anything we have :-)

mapiproxy/servers/default/asyncemsmdb/dcesrv_asyncemsmdb.c
```diff
@@ -806,10 +817,10 @@ static int process_newmail_notification(TALLOC_CTX *mem_ctx,
...
 				    struct tevent_fd *fde,
 				    uint16_t flags,
-				    void *private_data)
+				    void *asyncemsmdb_session)
 {
-	struct asyncemsmdb_private_data			*p = talloc_get_type(private_data,
-									     struct asyncemsmdb_private_data);
+	struct exchange_asyncemsmdb_session		*p = talloc_get_type(asyncemsmdb_session,
+									     struct exchange_asyncemsmdb_session);
 	TALLOC_CTX					*mem_ctx;
```
jkerihuel: Can you revert these variable renaming changes. These are imho cosmetic changes and it makes it more difficult for me to review the actual session change being discussed and its scope.

exekias: I think that this new name is way more easy to understand, and basically what all this change does is unify session handling with the rest of backends. Are you saying that you don't like the name or it's too difficult for you to review with both changes at the same time? I can split concerns into different commits

jkerihuel: No, I am kindly ask you to revert the variable naming change, that's all. I do not want to have a discussion about cosmetic changes or even be bothered by cosmetic changes in a PR that has impact on all openchange server and should require my full attention.

exekias: I have to say that I feel bad about the language here, I'm not trying to bother anyone but contributing to the project, and make a meaningful change that should improve it. I'm leaving this PR by the moment

PR #390, opened 2015-08-24
OpenChangeDB logger backend minor improvements + fixing compilation warnings

sixstone-qq:
* Implement missing unit tests for openchangedb_logger unittest.
* Avoid compilation warnings in openchangedb_multitenancy unittest + COPYRIGHT note
* Print out the change number returned value for `get_new_changeNumber` and `get_new_changeNumbers` for easing debugging purposes.

### DIFF
--- a/testsuite/libmapiproxy/openchangedb_logger.c
+++ b/testsuite/libmapiproxy/openchangedb_logger.c
@@ -12,7 +12,7 @@
 
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
    GNU General Public License for more details.
 
    You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


testsuite/libmapiproxy/openchangedb_logger.c
```diff
@@ -41,10 +44,26 @@
...
+        int get_fid;
+        int get_ReceiveFolder;
+};
+
+
 static TALLOC_CTX *mem_ctx;
 static struct openchangedb_context *oc_ctx;
-
-static struct openchangedb_context functions_called;
+static struct openchangedb_context_checker functions_called;
```
jkerihuel: What is the benefit of introducing this openchangedb_context_checker instead of keeping using openchangedb_context for functions_called?

sixstone-qq: Avoid compilation warnings in clang travis output. I think it is a hack to use a function pointer as an integer. IMHO this kind of hacks could lead to bugs and reduce the readability of the code. However, this has a clear drawback, we have to add every new unit tested procedure. If you have other solution to avoid that hack and change, I'll happily apply it. If you are umconfortable with this change, let me know as well.

jkerihuel: :+1: thanks for the highlights

jkerihuel:
Looks good to me. Any comments before merging?

blaxter:
:+1:

PR #391, opened 2015-08-25
Avoid race condition uploading changes which made new objects be missed

sixstone-qq:
This is done by returning the state (Change Number and Identifiers) of the involved objects on upload changes.

Futhermore, we return CnsetSeen and CnsetRead meta properties when a ImportReadStateChanges is done avoiding sync issues and it includes several useful debug messages.

There is more information in the commit messages.

This requires to work https://github.com/zentyal/sogo/pull/171 with SOGo backend.


mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -2569,6 +2580,11 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportDeletes(TALLOC_CTX *mem_ctx,
 
 	owner = emsmdbp_get_owner(synccontext_object);
 	openchangedb_get_MailboxReplica(emsmdbp_ctx->oc_ctx, owner, &repl_id, &replica_guid);
+	object_ids = talloc_zero_array(synccontext_object, uint64_t, object_array->cValues);
```
jkerihuel: where is object_array initialized?

sixstone-qq: good catch.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3164,6 +3200,17 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncOpenCollector(TALLOC_CTX *mem_ctx,
...
 	handles[mapi_repl->handle_idx] = synccontext_rec->handle;
 
+	/* Create an idset to store the involved fmids in the upload
+	   operations ids generated by these operations and current
+	   change number to retrieve only state after updating
+	   uploaded changes */
+	synccontext_object->object.synccontext->involved_fmids = RAWIDSET_make(synccontext_object, false, false);
+	retval = openchangedb_get_next_changeNumber(emsmdbp_ctx->oc_ctx, emsmdbp_get_owner(synccontext_object), &synccontext_object->object.synccontext->next_cn);
+	if (retval != MAPI_E_SUCCESS) {
+		mapi_repl->error_code = retval;
```
blaxter: I'd be lovely a `OC_DEBUG` message here.

blaxter:
Everything looks ok but `RopSyncImportDeletes` that can't work with the issue that Julien has commented. I'll perform also, as soon as I can, some manual testing.

sixstone-qq:
@blaxter in order to test ImportSyncDeletions, just remove a message from `Deleted items`. I'd like to implement client side code for that ROP in order to add a mapitest for it. Likewise, I'm in the process of proposing the ImportReadStateChanges client side code for another mapitest.

blaxter:
:+1: awesome fix

jkerihuel:
doing a last review

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -2570,12 +2581,20 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportDeletes(TALLOC_CTX *mem_ctx,
 	owner = emsmdbp_get_owner(synccontext_object);
 	openchangedb_get_MailboxReplica(emsmdbp_ctx->oc_ctx, owner, &repl_id, &replica_guid);
 
+	object_array = &request->PropertyValues.lpProps[0].value.MVbin;
```
jkerihuel: :hand: additional checks required. You assign a tainted data directly without prior checking if the value exists, if the index exists and if the property at this index is really of type PT_MV_BINARY:

* MS-OXCFXICS 2.2.3.2.4.5.1 states that PropertyValues (variable): An array of TaggedPropertyValue structures ([MS-OXCDATA] section 2.11.4). This value MUST NOT be NULL. -> we miss this check before assignation.

* The documentation is unclear to me whether we should expect a PidTag or just PT_MV_BINARY type and PidTag = 0x0, moreover the fixed position statement is not clear.  However as long as we have PT_MV_BINARY, you should be able to extract it using on of the convenience functions in libmapi/property.c

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -2570,12 +2581,20 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportDeletes(TALLOC_CTX *mem_ctx,
 	owner = emsmdbp_get_owner(synccontext_object);
 	openchangedb_get_MailboxReplica(emsmdbp_ctx->oc_ctx, owner, &repl_id, &replica_guid);
```
jkerihuel: I see you have added retval checks on openchangedb_get_MailboxReplica in another part of the patch, please propagate this change here too.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -2524,6 +2534,7 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportDeletes(TALLOC_CTX *mem_ctx,
 	struct GUID				replica_guid;
 	uint16_t				repl_id;
 	struct mapi_SBinaryArray		*object_array;
+	uint64_t				*object_ids;
```
jkerihuel: Initialize object_ids to NULL. If you apply the goto end: when checking openchangedb_get_mailboxReplica, it will lead to inconsistent behavior.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -2570,12 +2581,20 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportDeletes(TALLOC_CTX *mem_ctx,
...
+		goto end;
+	}
+
 	if (request->Flags & SyncImportDeletes_Hierarchy) {
-		object_array = &request->PropertyValues.lpProps[0].value.MVbin;
 		for (i = 0; i < object_array->cValues; i++) {
 			ret = oxcfxics_fmid_from_source_key(emsmdbp_ctx, owner, object_array->bin + i, &objectID);
 			if (ret == MAPISTORE_SUCCESS) {
 				emsmdbp_folder_delete(emsmdbp_ctx, synccontext_object->parent_object, objectID, 0xff);
+				object_ids[i] = objectID;
```
jkerihuel: We can end up with empty entries in object_ids if oxcfxics_fmid_from_source_key returns something different from MAPISTORE_SUCCESS. Is this the intended behavior?

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3164,6 +3202,18 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncOpenCollector(TALLOC_CTX *mem_ctx,
 	synccontext_object->object.synccontext->request.contents_mode = (mapi_req->u.mapi_SyncOpenCollector.IsContentsCollector != 0);
 	handles[mapi_repl->handle_idx] = synccontext_rec->handle;
 
+	/* Create an idset to store the involved fmids in the upload
+	   operations ids generated by these operations and current
+	   change number to retrieve only state after updating
+	   uploaded changes */
+	synccontext_object->object.synccontext->involved_fmids = RAWIDSET_make(synccontext_object, false, false);
```
jkerihuel: Add a check on RAWIDSET_make to check if involved_fmids is NULL or not. RAWIDSET_make DOES NOT check talloc retval at the moment but will when I finish review the libmapi/idset.c memory allocation. This will avoid one line patch. Thanks!

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3353,8 +3403,27 @@ _PUBLIC_ enum MAPISTATUS EcDoRpc_RopSyncImportReadStateChanges(TALLOC_CTX *mem_c
 
 	request = &mapi_req->u.mapi_SyncImportReadStateChanges;
 
+	retval = openchangedb_get_MailboxReplica(emsmdbp_ctx->oc_ctx, emsmdbp_get_owner(synccontext_object), NULL, &replica_guid);
+	if (retval != MAPI_E_SUCCESS) {
+		OC_DEBUG(5, "Impossible to get %s mailbox replica guid", emsmdbp_get_owner(synccontext_object));
+		mapi_repl->error_code = MAPI_E_CALL_FAILED;
```
jkerihuel: :question: returning MAPI_E_CALL_FAILED makes sense to me, but would be good if we could ensure that vs Outlook. Would you any way to test this easily and see if it causes Outlook to produce a sync issue or if it is OK with that and move on?

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3428,44 +3526,11 @@ static enum MAPISTATUS oxcfxics_fill_transfer_state_arrays(TALLOC_CTX *mem_ctx,
...
-	if (data_pointers && !retvals[0]) {
-		nr_eid = *(uint32_t *) data_pointers[0];
-		/* No content, return */
-		OPENCHANGE_RETVAL_IF(!nr_eid, MAPI_E_SUCCESS, local_mem_ctx);
-	} else {
-		OC_DEBUG(1, "ERROR: could not retrieve number of rows in table\n");
-		talloc_free(local_mem_ctx);
-		return MAPI_E_CALL_FAILED;
-	}
-
```
jkerihuel: :question: I don't understand from current patch or commit why this code was removed. Can you give some context/explanations why it was needed?

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3480,27 +3545,58 @@ static enum MAPISTATUS oxcfxics_fill_transfer_state_arrays(TALLOC_CTX *mem_ctx,
...
+                                 table_object->object.table->denominator, synccontext->next_cn);
+
+			OPENCHANGE_RETVAL_IF(table_object->object.table->denominator == 0,
+					     MAPI_E_SUCCESS, local_mem_ctx);
+		}
 	}
+
+	if (synccontext->request.is_collector) {
+		involved_fmids = RAWIDSET_convert_to_idset(local_mem_ctx, synccontext->involved_fmids);
+		IDSET_dump(involved_fmids, "involved fmids");
```
jkerihuel: It this intended to remain here? 

IDSET_dump prints with OC_DEBUG level=0. I don't know how often this operation is called, but so far we have tried to limit the scope of use of level=0.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3480,27 +3545,58 @@ static enum MAPISTATUS oxcfxics_fill_transfer_state_arrays(TALLOC_CTX *mem_ctx,
...
+				OC_DEBUG(5, "Skipping putting eid (0x%.16"PRIx64") as it is not included in an upload operation", eid);
+				continue;
+			}
 			if (retvals[sync_data->prop_index.change_number]) {
 				OC_DEBUG(1, "mandatory property PidTagChangeNumber not returned for message %.16"PRIx64, eid);
 				talloc_free(local_mem_ctx);
 				return MAPI_E_CALL_FAILED;
 			}
+
+			emsmdbp_replid_to_guid(emsmdbp_ctx, owner, eid & 0xffff, &replica_guid);
```
jkerihuel: Check retval, it can return MAPI_E_NOT_FOUND and replica_guid to be junk because not memset/initialized beforehand. 

Additionally a memset on replica_guid to initialize its content to 0 before hand wouldn't be harmful.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3480,27 +3545,58 @@ static enum MAPISTATUS oxcfxics_fill_transfer_state_arrays(TALLOC_CTX *mem_ctx,
...
+			RAWIDSET_push_guid_glob(sync_data->eid_set, &replica_guid, (eid >> 16) & 0x0000ffffffffffff);
```
jkerihuel: Note to myself: probably a mem_leak in RAWIDSET_push_guid_glob with the talloc_zero on zero_guid which is not released anywhere.

mapiproxy/servers/default/emsmdb/oxcfxics.c
```diff
@@ -3480,27 +3545,58 @@ static enum MAPISTATUS oxcfxics_fill_transfer_state_arrays(TALLOC_CTX *mem_ctx,
...
+			OC_DEBUG(5, "Number of updated messages to return: %d with CN >= 0x%.16"PRIx64,
+                                 table_object->object.table->denominator, synccontext->next_cn);
+
+			OPENCHANGE_RETVAL_IF(table_object->object.table->denominator == 0,
+					     MAPI_E_SUCCESS, local_mem_ctx);
+		}
 	}
+
+	if (synccontext->request.is_collector) {
+		involved_fmids = RAWIDSET_convert_to_idset(local_mem_ctx, synccontext->involved_fmids);
```
jkerihuel: Following the logic of RAWIDSET_convert_to_idset and subsequent calls of this function in this file, talloc_free should be called on involved_fmids. Check for example https://github.com/openchange/openchange/pull/391/files#diff-382a715683da99a802a331f231eade7bL3590

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: I think this is out of the scope of this PR. I've only added to the one that it was required (no function was used). I'm pretty ok to introduce that change but in another PR. My reasoning is to perform the tests against those changes and having branches in git is cheap and PRs even more. If you are unconfortable with that reasoning, I will modify this PR with those changes but with new commits.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: Here, it applies even more than above. In this PR I just move the code a few lines earlier.

If we want to make sure any other change does not break things, I would suggest to put off the additional checks to other pull request. Indeed, we have to add mapitest and client side ImportDeletes operation to make sure everything works as expected :)

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: You can test it without a problem following these steps:

* Modify this code to return always MAPI_E_CALL_FAILED
* Mark a message as unread
* Sync the folder

According to spec, it will try it using Online mode (SetMessageReadFlag ROP), so no sync issues are expected. Indeed, in Outlook 2010 it tries both at the same time.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: It wasn't needed because it seems an early optimisations on previous implementation to avoid going through an empty table. This is not necessary and it reduces a lot the readability of the procedure IHMO. If there is no content in the end, no new change numbers neither identifiers will be added to the returned ids. What do you think?

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: I'm against that logic. Indeed, I haven't changed previous logic because I didn't want to make this kind of changes without a full wipe-out. What's the benefit of freeing here the idsets? IMHO, talloc local contexts do the job perfectly to dynamic allocate variables for a function at the beginning and, then, at the end freeing them. Am I nuts?

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: Yes. Indeed, I have a stash to modify IDSET_dump to log level 5. It makes no sense to output at 0. No user will ever understand this. It takes 4 weeks to learn it properly, so developer intended.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: Good catch!

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: Np, I'll add it.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: it won't hurt anyway.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: See https://github.com/openchange/openchange/pull/394

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: I've done so in #397 , let me know if I miss something.

mapiproxy/servers/default/emsmdb/oxcfxics.c
sixstone-qq: Done in #398 .

mapiproxy/servers/default/emsmdb/oxcfxics.c
blaxter: In the example linked those idset are created out of a memory context so they need to be freed explicitly, so I this is ok how is done here

blaxter:
I think all concerns have been either fixed or moved to another pull request, so let's merge this to move forward.

PR #393, opened 2015-08-26
rpcproxy: Fixes on sad paths

sixstone-qq:
* Manage client disconnections waiting for CONN/B1 RTS packet 
* Check properly how to wait for IN channel from Outbound proxy

### DIFF
--- a/mapiproxy/services/web/rpcproxy/rpcproxy/channels.py
+++ b/mapiproxy/services/web/rpcproxy/rpcproxy/channels.py
@@ -280,8 +280,18 @@ class RPCProxyInboundChannelHandler(RPCProxyChannelHandler):
             self.logger.debug("processing IN channel request")
 
             self.client_socket = environ["wsgi.input"]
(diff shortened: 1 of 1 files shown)


mapiproxy/services/web/rpcproxy/rpcproxy/channels.py
```diff
@@ -447,8 +457,7 @@ def _wait_IN_channel(self):
 
             connected = True
         else:
-            self.logger.info("IN channel failed to connect in the last"
-                             " 10 seconds")
+            self.logger.info("Failed to connect to IN channel in last 10 seconds")
```
exekias: I think the previous message was more correct, you are connected to the IN channel, but waiting for it to connect to samba

exekias:
:+1:, as we saw this is a strange corner case but this code should handle that

blaxter:
:+1: merging and adding an entry in `CHANGES.md` file, something like: `rpcproxy handles better disconnections`.

PR #394, opened 2015-08-26
libmapi: IDSET_dump dumps to log level 5

sixstone-qq:
It makes no sense to use log level 0 when no user will understand
that structure without reading the oxcfxics document. What I'm sure
not even all OC devs has read throughfully. :)


blaxter:
:+1:

PR #397, opened 2015-09-06
Improvements over SyncImportDeletes OXCFXICS ROP

sixstone-qq:
- Implement SyncImportDeletes operation in client side
  - Along with mapitest to test it
- Folder deletion on cached mode now works
  - It was only working if the folder was at INBOX's level
- Added additional checks on SyncImportDeletes
  - As suggested in #391 comment

It is based #391 to avoid later conflict resolution.


blaxter:
When I say re, you say base. re...

PR #398, opened 2015-09-06
oxcfxics: Checks on openchangedb_get_MailboxReplica

sixstone-qq:
As suggested jkerihuel comment in #391 and it is based on that code base.

There are places where the check has not been added because the returned value is not checked and I think this work was already started.


blaxter:
rebase

KTHXBYE

jkerihuel:
Looks good.

Are Outlook Tests and Zinc Tests failures related or not?

sixstone-qq:
@jkerihuel No, there are not. The tests are unstable right now.

PR #400, opened 2015-09-10
ndr_fasttransfer: avoid infinite loop

blaxter:
When `ndr_pull_uint32` failed on `ndr_parse_FastTransferElement` we were ignoring the error (offset > 0) and moving on: infinite loop

### DIFF
--- a/ndr_fasttransfer.c
+++ b/ndr_fasttransfer.c
@@ -519,7 +519,7 @@ static int ndr_parse_propValue(TALLOC_CTX *mem_ctx, struct ndr_print *ndr,
    \param ndr pointer to the ndr_print structure
    \param ndr_pull pointer to the ndr_pull structure representing the fxbuffer
 
-   \return value >= 0 on success, otherwise -1
+   \return 0 on success, otherwise -1 or > 0
 */
 static int ndr_parse_FastTransferElement(TALLOC_CTX *mem_ctx, struct ndr_print *ndr, struct ndr_pull *ndr_pull)
 {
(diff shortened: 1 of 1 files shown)


ndr_fasttransfer.c
```diff
@@ -572,15 +571,14 @@ static int ndr_parse_FastTransferStream(struct ndr_print *ndr, DATA_BLOB buf)
 	ndr_set_flags(&ndr_pull->flags, LIBNDR_FLAG_NOALIGN);
 	ndr_set_flags(&ndr->flags, LIBNDR_FLAG_NOALIGN);
 
-	while (ndr_pull->offset < ndr_pull->data_size -1) {
-		offset = ndr_parse_FastTransferElement(mem_ctx, ndr, ndr_pull);
-		if (offset < 0) {
+	while (ndr_pull->offset < ndr_pull->data_size - 1) {
+		ret = ndr_parse_FastTransferElement(mem_ctx, ndr, ndr_pull);
+		if (ret != 0) {
```
sixstone-qq: The documentation on `ndr_parse_FastTransferElement` says that `ret` > 0 is ok, not an error. If the documentation is wrong, I'd suggest to modify it. If it is right, please change the check :).

jkerihuel:
Indeed, there is an infinite loop here, good catch. 

I am however not OK with the fix. I will come up with a different approach today, that we may eventually apply elsewhere if similar issue arise.

Keeping open for now, until new patch is proposed.

jkerihuel:
Do you have the sample that highlights the infinite loop?

Could be useful to write a test suite for ndr_fastrransfer and avoid future regression.

blaxter:
No I don't

PR #401, opened 2015-09-11
Fix error checking in the FastTransfer buffer parse.

jkerihuel:
Make it compliant with the model: error if < 0, otherwise success.

### DIFF
--- a/ndr_fasttransfer.c
+++ b/ndr_fasttransfer.c
@@ -574,7 +574,7 @@ static int ndr_parse_FastTransferStream(struct ndr_print *ndr, DATA_BLOB buf)
 
 	/* Map DATA_BLOB to ndr pull context */
 	ndr_pull = talloc_zero(mem_ctx, struct ndr_pull);
-	NDR_ERR_HAVE_NO_MEMORY(ndr_pull);
+	if (!ndr_pull) return -1;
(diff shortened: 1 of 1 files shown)


blaxter:
So instead of using the same error model of ndr layer, we have a different model (with different macros) only for this file. When I found the infinite loop actually my first idea was to replace all the return codes to `enum ndr_err_code` as that's the best solution but as this's not really important (500lines code) I just created the pull request with the fast and easy to understand solution.

For other changes I'd appreciate if instead of redoing my patches, you just comment your pov in the original pr to discuss it.

jkerihuel:
It would not make sense to rely on ndr_err_code and unify error code to this regards in ndr_fasttransfer file. We do not make any functional use of NDR layer error code, we do not raise it to upper layers and more importantly we do not want to raise any specific error code upon failure: ndr_fasttransfer is a developer convenience. We want to decode it fully and if we can't or hit a decoding issue, we want to display a binary blob instead, then return success. Eventually inspect it later but in any case developer's inspection is required.

Hence the current function signature makes more sense to me: -1 on error, 0 on success, period. ndr_err_code for failure is > 0, which would require to update the complete logic for no benefits.

blaxter:
Well, it doesn't make sense to me but I don't care, merge whatever you want.

PR #405, opened 2015-11-09
Log failure opening the backend store path

roelvanmeer:
Log a message if the backend store path cannot be opened. If the default
backend path is used, and it does not exist, you currently only get the
messages:

  (mapistore_backend_init): No mapistore backends available (using backend path '(null)'
  (mapistore_init): mapistore_backend_init: Storage backend initialization failed
  (emsmdbp_init): MAPISTORE initialization failed

It seems helpful to log exactly what goes wrong.
This would have saved me some time debugging.


sixstone-qq:
:+1:

PR #410, opened 2015-12-16
Fix crash caused by empty named_properties table

roelvanmeer:
If the named_properties table is empty, the query
  SELECT max(mappedId) FROM <table>
returns NULL, causing a crash in strtol() later.
Checking if row[0] is not NULL makes the code return MAPISTORE_ERR_DATABASE_OPS instead of crashing.


sixstone-qq:
:+1:

jkerihuel:
applied, thx @roelvanmeer!

PR #424, opened 2018-03-07
Fix build with new libical

mastersin:


jelmer:
This branch seems to include more than the described changes. It seems to be against the zentyal openchange repo?

